PR #3, opened 2019-04-16
Created operator to configure Alertmanager with PagerDuty and DMS

dak1n1:
The controller will automatically keep the alertmanager config up-to-date based on the presence or absence of secrets for Pager Duty and Dead Man's Snitch.

It's in a working state, but lacks the ability to reload alertmanager in order for the config changes to take effect. Also lacking unit tests and metrics. I've opened issues to address those items:

https://github.com/openshift/configure-alertmanager-operator/issues/1
https://github.com/openshift/configure-alertmanager-operator/issues/2

This the first two commits contain a lot of generated code. The 3rd commit contains changes specific to this operator.


pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,428 @@
...
+func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling Secret")
+
+	// This operator is only interested in the 3 secrets listed below. Skip reconciling for all other secrets.
+	if request.Name != "alertmanager-main" && request.Name != "dms-secret" && request.Name != "pd-secret" {
+		reqLogger.Info("Skip reconcile: No changes detected to alertmanager secrets.")
+		return reconcile.Result{}, nil
+	}
+	fmt.Println("DEBUG: Started reconcile loop")
```
wshearn: Why are you using fmt.Println instead of the logger? reqLogger

lisa:
https://github.com/dak1n1/configure-alertmanager-operator/pull/3 addresses some things:

* stray `)` in `config/config.go`
* use new-new Makefile pattern
* use openshift/release Dockerfile pattern

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,428 @@
...
+// removeFromReceivers removes the specified index from a slice of Receivers.
+func removeFromReceivers(slice []*alertmanager.Receiver, i int) []*alertmanager.Receiver {
+	// copy all the elements in the slice except for the one we want to delete
+	slicecopy := []*alertmanager.Receiver{}
+	for index, receiver := range slice {
+		if index != i {
+			fmt.Println("DEBUG: The index", index, "doesn't match the element we want to delete:", i, ". So copy it over to slicecopy")
+			slicecopy = append(slicecopy, receiver)
+			fmt.Println("DEBUG: contents of slicecopy so far:")
+			for _, x := range slicecopy {
```
rafael-azevedo: I don't think you need this function here It just prints out the data and isn't modifying anything. Is there a reason this needs to be logged ? Also if you want to delete an item from an array it is easier to do something like this 

```
a = append(a[:i], a[i+1:]...) ---> This creates a copy of the array from 0:i and from i+:(end)
// or
a = a[:i+copy(a[i:], a[i+1:])]
```

where i is the index you want to delete. 

Let me know if you have any questions

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,428 @@
...
+				fmt.Println(x)
+			}
+		}
+	}
+	return slicecopy
+}
+
+// removeFromRoutes removes the specified index from a slice of Routes.
+func removeFromRoutes(slice []*alertmanager.Route, i int) []*alertmanager.Route {
```
rafael-azevedo: Same issues here as noted above

pkg/types/alertmanagerconfig.go
```diff
@@ -0,0 +1,126 @@
...
+func (c Config) String() string {
+	b, err := yaml.Marshal(c)
+	if err != nil {
+		return fmt.Sprintf("<error creating config string: %s>", err)
+	}
+	return string(b)
+}
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface for Config.
+func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
rafael-azevedo: Why are you implementing the unmarshal function instead of using the built in ?

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,428 @@
...
+// removeConfigFromAlertManager removes a Receiver config and the associated Route from Alertmanager.
+// The changes are kept in memory until committed using function updateAlertManagerConfig().
+func removeConfigFromAlertManager(r *ReconcileSecret, request *reconcile.Request, amconfig *alertmanager.Config, receivername string) {
+	fmt.Println("DEBUG: Checking for receiver", receivername, "in Alertmanager config")
+	for i, receiver := range amconfig.Receivers {
+		fmt.Println("DEBUG: Found Receiver named:", receiver.Name, "Checking against receivername:", receivername)
+		if receiver.Name == receivername {
+			fmt.Println("DEBUG: Deleting receiver named:", receiver.Name)
+			amconfig.Receivers = removeFromReceivers(amconfig.Receivers, i)
+		} else {
```
rafael-azevedo: I don't think you need the else in here

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,428 @@
...
+			fmt.Println("DEBUG: Skipping Receiver named", receiver.Name)
+		}
+	}
+
+	for i, route := range amconfig.Route.Routes {
+		fmt.Println("DEBUG: Found Route for reeceiver named:", route.Receiver, "Checking against receivername:", receivername)
+		if route.Receiver == receivername {
+			fmt.Println("DEBUG: Deleting Route for Receiver:", route.Receiver)
+			amconfig.Route.Routes = removeFromRoutes(amconfig.Route.Routes, i)
+		} else {
```
rafael-azevedo: I also think else can be removed here

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,428 @@
...
+	reqLogger.Info("Reconciling Secret")
+
+	// This operator is only interested in the 3 secrets listed below. Skip reconciling for all other secrets.
+	if request.Name != "alertmanager-main" && request.Name != "dms-secret" && request.Name != "pd-secret" {
+		reqLogger.Info("Skip reconcile: No changes detected to alertmanager secrets.")
+		return reconcile.Result{}, nil
+	}
+	fmt.Println("DEBUG: Started reconcile loop")
+
+	// This block looks at the specific instance of Secret for each Secret in the `openshift-monitoring` namespace.
```
rogbas: Do we want to monitor secrets on the `openshift-monitoring` ns as stated in that comment?

I noticed in the operator Deployment (06_operator.yaml file) that the WATCH_NAMESPACE parameter is set to the operator's ns (metadata.namespace), so the way it is it will watch for secrets only on the `configure-alertmanager-operator` ns

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,428 @@
...
+				alertmanagerconfig := getAlertManagerConfig(r, &request)
+				fmt.Println("INFO: Dead Man's Snitch secret is absent. Removing Watchdog config from Alertmanager")
+				removeConfigFromAlertManager(r, &request, &alertmanagerconfig, "watchdog")
+				updateAlertManagerConfig(r, &request, &alertmanagerconfig)
+				return reconcile.Result{}, nil
+			}
+		} else {
+			// Error and requeue in all other circumstances.
+			// Don't requeue if a Secret is not found. It's valid to have an absent Pager Duty or DMS secret.
+			fmt.Println("DEBUG: error reading object. Requeuing request")
```
rogbas: I would switch those Println by logger calls

config/config.go
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+const (
+	OperatorName          string = "configure-alertmanager-operator"
+	OperatorNamespace     string = "configure-alertmanager-operator"
+			)
```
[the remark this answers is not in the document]
lisa: See https://github.com/dak1n1/configure-alertmanager-operator/pull/3, which addresses this.

build/bin/entrypoint
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -e
+
```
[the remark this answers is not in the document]
lisa: See https://github.com/dak1n1/configure-alertmanager-operator/pull/3, which addresses this.

build/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+COPY --from=build ${OPERATOR_PATH}/build/_output/bin/${OPERATOR_BIN} /usr/local/bin/${OPERATOR_BIN}
+USER ${USER_UID}
+LABEL io.openshift.managed.name="configure-alertmanager-operator" \
+      io.openshift.managed.description="Operator to configure Alertmanager with PagerDuty and Dead Man's Snitch."
+
+COPY build/bin /usr/local/bin
+
+ENTRYPOINT ["/usr/local/bin/configure-alertmanager-operator"]
+
+USER ${USER_UID}
```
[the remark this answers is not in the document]
lisa: See https://github.com/dak1n1/configure-alertmanager-operator/pull/3, which addresses this.

build/bin/entrypoint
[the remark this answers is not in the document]
rogbas: Thanks for pointing it out

pkg/controller/secret/secret_controller.go
dak1n1: Because I found it to be a lot easier to read when I'm debugging things. Otherwise I have to scroll past a lot of output and it's harder on human eyes to read through the json formatting. Here's some example output. It helps separate out times when the operator is processing things that don't matter, vs things that a human would care about:

https://gist.githubusercontent.com/dak1n1/24a90c2ce8cfcd5f90021d05c8d70dca/raw/6027eb5ceab698557ac00e39e3aad53fe39878ef/gistfile1.txt

pkg/controller/secret/secret_controller.go
dak1n1: Thanks, that was some extra debug output that I don't need anymore. I was having trouble deleting Receivers/Routes from the Alertmanager config. I'll remove that.

dak1n1: Ok, this is fixed and tested now.

pkg/types/alertmanagerconfig.go
dak1n1: This is something I copied from the Alertmanager Config data type, from an upstream repo. If you unmarshal an Alertmanager Config using the default unmarshaler, it gives you a bunch of memory addresses instead of valid yaml. For example, this is how the list of receivers is printed out:

```
[0xc4201831d0 0xc4201832c0 0xc4201833b0]
```

And individual receivers look like this:
```
&{pagerduty [] [0xc0009220e0] [] [] [] [] [] [] []}
```

It also helps to ensure that the list of receivers each has a unique name, which is required for a valid Alertmanager config.

pkg/controller/secret/secret_controller.go
dak1n1: More debug output. I removed it.

deploy/00_namespace.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
dak1n1: Thanks, I went ahead and moved the `deploy` dir to `manifests`.

pkg/controller/secret/secret_controller.go
dak1n1: Good catch! WATCH_NAMESPACE in this file is misleading and has no effect, because it's actually set here instead:

https://github.com/dak1n1/configure-alertmanager-operator/blob/initial_create/cmd/manager/main.go#L79

Should I configure that differently? For example, in the dedicated-admin-operator, the value `namespace` is used there instead:

https://github.com/openshift/dedicated-admin-operator/blob/master/cmd/manager/main.go#L89

But if I do that, it'll watch the namespace `configure-alertmanager-operator` because my operator pods run there. I think I'd better move my operator to run completely inside the `openshift-monitoring` namespace instead. That way I can configure mine like the dedicated-admin-operator. Does that sound ok?

jewzaam: Sounds fine to me to move to `openshift-monitoring`.  Think we talked about this possibility a while back.  And if the secrets are synced to that NS as well then this operator is confined to a single NS.

dak1n1: Ok, it's working in the openshift-monitoring namespace now. Thanks!

pkg/controller/secret/secret_controller.go
dak1n1: I'll go ahead and make everything use log instead. Maybe I can configure the timestamps to make more sense, and clean up the amount of stuff logger is printing to make it easier to read.

dak1n1: I opened an RFE to make the timestamps easier to read. I'll check it out after this PR merges. https://github.com/openshift/configure-alertmanager-operator/issues/4

build/bin/entrypoint
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -e
```
lisa: This file isn't needed, and can be deleted.

build/bin/user_setup
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
lisa: This file isn't needed, and can be deleted.

build/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+WORKDIR ${OPERATOR_PATH}
+RUN make go-build
+
+FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+ENV OPERATOR_PATH=/go/src/github.com/openshift/configure-alertmanager-operator \
+￼     OPERATOR_BIN=configure-alertmanager-operator
+
+WORKDIR /root/
+COPY --from=builder ${OPERATOR_PATH}/build/_output/bin/${OPERATOR_BIN} /usr/local/bin/${OPERATOR_BIN}
+USER ${USER_UID}
```
lisa: Specifying the `USER` isn't needed, and the `${USER_UID}` variable isn't defined.

Makefile
```diff
@@ -0,0 +1,46 @@
+include project.mk
```
lisa: Please use the standard format used by all other Makefiles: See https://github.com/lisa/configure-alertmanager-operator/blob/a4483abaffc9fb0acedf07992897294ca1a1297c/Makefile for reference. Customizations can be made after the mark.

jewzaam: Would it help to have a PR into this fork to apply the standard bits?  Otherwise might suggest we tackle that in a separate PR and not mix it into this initial merge.  It's important, but so is getting a good baseline into `master`.

lisa: I did, https://github.com/dak1n1/configure-alertmanager-operator/pull/3, but it was closed.

pkg/controller/secret/secret_controller.go
dak1n1: Done

Gopkg.toml
```diff
@@ -0,0 +1,75 @@
...
+  version = "=v0.26.0"
+
+[[override]]
+  name = "sigs.k8s.io/controller-runtime"
+  version = "=v0.1.10"
+
+[[constraint]]
+  name = "github.com/operator-framework/operator-sdk"
+  # The version rule is used for a specific release and the master branch for in between releases.
+  branch = "master" #osdk_branch_annotation
```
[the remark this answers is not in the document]
dak1n1: Thanks for the advice! I reinstalled operator-sdk at version `v0.7.0` and updated the dependencies like you suggested. I did a fresh build/deployment of the operator and confirmed it still works as expected.

build/bin/entrypoint
dak1n1: Done

build/bin/user_setup
dak1n1: Done

build/Dockerfile
dak1n1: I had to revert to my original docker file to make the build work. `USER_UID=1001` is set

Makefile
dak1n1: I'd prefer to get this initial commit in before having other contributors jump in. I tried merging Lisa's changes into my branch, but with all the merge conflicts, it was hard to get the changes in and hard to know which parts were broken. I had added them to this PR, but I had to revert the changes to the Docker file because I couldn't get it to build. It's too many moving parts. I'll just get this PR merged ASAP for a baseline so others can begin contributing to the repo.

dak1n1: Added standard make file.

PR #5, opened 2019-04-18
SREP-1143: Update Dockerfile to address build failures

lisa:
Signed-off-by: Lisa Seelye <lseelye@redhat.com>


jewzaam:
Recommend update to build in `quay.io/${USER}` by default, per standard make stuff

PR #16, opened 2019-05-17
SREP-1186: Add olm artifact template

lisa:
See https://github.com/app-sre/saas-hive/pull/24 (I'll update the path on that PR)

### DIFF
--- a/build/templates/olm-artifacts-template.yaml.tmpl
+++ b/build/templates/olm-artifacts-template.yaml.tmpl
@@ -41,25 +41,6 @@ objects:
           mediatype: ""
         publisher: Red Hat
         sourceType: grpc
-- apiVersion: hive.openshift.io/v1alpha1
-  kind: SelectorSyncSet
-  metadata:
-    generation: 1
-    name: osd-configure-alertmanager-operator
-  spec:
-    clusterDeploymentSelector:
-      matchLabels:
-        api.openshift.com/managed: "true"
-    resourceApplyMode: sync
-    resources:
-    - apiVersion: v1
-      kind: Namespace
-      metadata:
-        name: openshift-monitoring
-      annotations:
-        openshift.io/node-selector: ""
-      labels:
-        openshift.io/cluster-monitoring: "true"
     - apiVersion: operators.coreos.com/v1alpha1
       kind: Subscription
       metadata:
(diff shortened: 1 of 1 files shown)


build/templates/olm-artifacts-template.yaml.tmpl
```diff
@@ -0,0 +1,72 @@
...
+        openshift.io/node-selector: ""
+      labels:
+        openshift.io/cluster-monitoring: "true"
+    - apiVersion: operators.coreos.com/v1alpha1
+      kind: Subscription
+      metadata:
+        name: configure-alertmanager-operator
+        namespace: openshift-monitoring
+      spec:
+        channel: master
```
rogbas: Shouldn't master be replaced by the CHANNEL var here?

build/templates/olm-artifacts-template.yaml.tmpl
```diff
@@ -0,0 +1,72 @@
...
+    - apiVersion: operators.coreos.com/v1alpha1
+      kind: CatalogSource
+      metadata:
+        labels:
+          opsrc-datastore: "true"
+          opsrc-provider: redhat
+        name: configure-alertmanager-operator-registry
+        namespace: openshift-operator-lifecycle-manager
+      spec:
+        image: ${REGISTRY_IMG}:${CHANNEL}-${IMAGE_TAG}
```
rogbas: REGISTRY_IMG will come with ORG/IMG_NAME, correct?

jmelis: in other repos it does

build/templates/olm-artifacts-template.yaml.tmpl
jmelis: good catch

build/templates/olm-artifacts-template.yaml.tmpl
```diff
@@ -0,0 +1,53 @@
...
+- name: IMAGE_TAG
+  value: latest
+  required: true
+
+objects:
+- apiVersion: hive.openshift.io/v1alpha1
+  kind: SelectorSyncSet
+  metadata:
+    generation: 1
+    name: configure-alertmanager-operator-catalogsource
```
rogbas: Since it's a single SSS it doesn't make sense to have CatalogSource in the name anymore.
Maybe just `configure-alertmanager-operator` ?

lisa: WFM

PR #17, opened 2019-05-17
Deploy needed objects for prometheus metrics when the operator comes up

fahlmant:
This PR ensures that when the operator is deployed, all the proper objects are placed so prometheus can scrape the custom metrics.

-The operator creates a configure-alertmanager-operator service to expose the metrics over on deploy
-The operator creates a configure-alertmanager-operator servicemonitor to tell prometheus how to digest the metrics
-Removes the operator-sdk metrics service for now
-Includes a prometheus-k8s role for the configure-alertmanager-operator namespace that has permissions to get, list and watch services, pods and endpoints.
-Includes a prometheus-k8s rolebinding to allow the prometheus-k8s serviceaccount from the openshift-monitoring namespace to use the aformentioned role.

When the operator and the role/rolebindings are deployed, the configure-alertmanager-operator namespace (or wherever its running) must be given the following label to get everything working: openshift.io/cluster-monitoring="true"

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -9,11 +9,13 @@ import (
 
 	"github.com/openshift/configure-alertmanager-operator/pkg/apis"
 	"github.com/openshift/configure-alertmanager-operator/pkg/controller"
+	operatormetrics "github.com/openshift/configure-alertmanager-operator/pkg/metrics"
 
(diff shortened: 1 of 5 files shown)


dak1n1:
For reference, this will help solve https://github.com/openshift/configure-alertmanager-operator/issues/1

PR #18, opened 2019-05-20
Add metrics to secret controller

dak1n1:
Add metrics that can be scraped by Prometheus. Specifically, give the data needed in order for Prometheus to:

* Alert when `dms-secret` exists, but the corresponding AlertManager config does not exist.
* Alert when `pd-secret` exists, but the corresponding AlertManager config does not exist.
* Alert when Pager Duty config exists, but the corresponding secret (`pd-secret`) does not exist.
* Alert when Dead Man's Snitch config exists, but the corresponding secret (`dms-secret`) does not exist.

This PR includes a cherry-picked commit from https://github.com/openshift/configure-alertmanager-operator/pull/14, since I couldn't continue without the secret_controller.go changes included in there.


pkg/metrics/metrics.go
```diff
@@ -56,8 +79,50 @@ func RegisterMetrics() error {
...
-func UpdatePlaceholderGauge() {
+// UpdateSecretsMetrics updates all metrics related to the existance and contents of Secrets
+// used by configure-alertmanager-operator.
+func UpdateSecretsMetrics(list *corev1.SecretList, amconfig alertmanager.Config) {
+
+	// Default to false.
+	pdSecretExists := false
+	dmsSecretExists := false
+	amSecretExists := false
+	metricPDSecretExists.With(prometheus.Labels{"name": config.OperatorName}).Set(float64(0))
```
fahlmant: You set this metric potentially twice in this function: here and line 101. I would only set this (and other metrics) one time per run.

cmd/manager/main.go
```diff
@@ -129,6 +136,7 @@ func main() {
 	}
 
 	log.Info("Starting the Cmd.")
```
fahlmant: I misspoke, this is what this block should look like

```
	log.Info("Starting prometheus metrics")
	operatormetrics.StartMetrics()

	log.Info("Starting the Cmd.")

	// Start the Cmd
```

pkg/metrics/metrics.go
dak1n1: Yeah, that was intentional with all 5 metrics since I was defaulting them to false. But check out the latest commit and see if you like that better. I changed it to only update each metrics once per run.

dak1n1:
Thanks, I'll wait for https://github.com/openshift/configure-alertmanager-operator/pull/14/files to merge first, since this one depends on that.

fahlmant:
@dak1n1 So I'm looking at this again and you're now only updating if they're true. What happens if those bools are false?

dak1n1:
Yeah, that's why I had them set to 0 initially and updated to 1 conditionally. But I got rid of that and attempted to do as you asked by only setting them once. I must have misunderstood your request. But I've updated it to conditionally set to 0 now too.

pkg/controller/secret/secret_controller.go
```diff
@@ -253,7 +260,7 @@ func addPDSecretToAlertManagerConfig(r *ReconcileSecret, request *reconcile.Requ
 			"severity",
 		},
 		MatchRE: map[string]string{
-			"namespace": "^openshift$|openshift-.*|default$|kube$|kube-.*|logging$",
+			"namespace": "^kube-.*,^openshift-.*,^logging$,^default$,^openshift$,^ops-health-monitoring$,^ops-project-operation-check$,^management-infra$",
```
jewzaam: The following namespaces will not be created by SRE in OSD v4:

- ops-health-monitoring
- ops-project-operation-check
- management-infra

But, we blacklist these with dedicated-admin operator:
https://github.com/openshift/dedicated-admin-operator/blob/master/pkg/dedicatedadmin/dedicatedadmin.go#L53

So there is no risk unless we change the regex for dedicated-admin operator.

dak1n1:
Rebased and ready for final review.

pkg/controller/secret/secret_controller_test.go
```diff
@@ -188,7 +188,7 @@ func Test_addPDSecretToAlertManagerConfig(t *testing.T) {
 			"severity",
 		},
 		MatchRE: map[string]string{
-			"namespace": "^openshift$|openshift-.*|default$|kube$|kube-.*|logging$",
+			"namespace": "^kube-.*,^openshift-.*,^logging$,^default$,^openshift$,^ops-health-monitoring$,^ops-project-operation-check$,^management-infra$",
```
jewzaam: Since this is set in two places and it's pretty complex I'd prefer seeing it as a constant, but given this isn't owned by this operator I'll let it slide.  Unit test failures will catch changes out of sync?

dak1n1: Yeah, the unit test caught the mismatch this time. I could make it a constant and export it to the packages that need it.

dak1n1: Do you want me to do that on this PR?

jewzaam: Since you need some more eyes on it for a lgtm label, might as well.

standard.mk
```diff
@@ -29,7 +29,7 @@ MAINPACKAGE=./cmd/manager
 GOENV=GOOS=linux GOARCH=amd64 CGO_ENABLED=0
 GOFLAGS=-gcflags="all=-trimpath=${GOPATH}" -asmflags="all=-trimpath=${GOPATH}"
 
-TESTTARGETS := $(shell go list -e ./... | egrep -v "/(vendor)/")
+TESTTARGETS := ./pkg/controller/secret
```
lisa: Why hardcode it to this? the `go list ...` stuff will find all of the packages, and if they have no tests `make gotest` will skip, and that' ok.

manifests/04_operator.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       serviceAccountName: configure-alertmanager-operator
       containers:
         - name: configure-alertmanager-operator
-          image: quay.io/openshift-sre/configure-alertmanager-operator:v0.1.9-fc79391f
+          image: quay.io/openshift-sre/configure-alertmanager-operator:latest
```
lisa:
```suggestion
          image: quay.io/app-sre/configure-alertmanager-operator:staging-latest
```

These now live in `app-sre` quay.io organization. I also suggested a change from `latest` to `staging-latest` since there's no longer a `latest` tag.

standard.mk
dak1n1: I was getting docker build failures during the test phase, so I thought I needed to change this. But I just tested it out with the old code and it works now. I must have been hitting a different problem.

manifests/04_operator.yaml
dak1n1: ok, fixed.

PR #20, opened 2019-05-20
Build should push images to Quay

jharrington22:
This PR resolves an issue where the build pipeline fails because images are not pushed to Quay.

### DIFF
--- a/build/deploy.sh
+++ b/build/deploy.sh
@@ -10,4 +10,4 @@ fi
 
 # Build the image
 
-make IMAGE_REPOSITORY=$IMAGE_REPOSITORY build skopeo-push
+make IMAGE_REPOSITORY=$IMAGE_REPOSITORY build skopeo-push build-catalog-image
--- a/build/deploy.sh
+++ b/build/deploy.sh
@@ -10,4 +10,4 @@ fi
 
 # Build the image
 
-make IMAGE_REPOSITORY=$IMAGE_REPOSITORY build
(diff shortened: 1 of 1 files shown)


jharrington22:
@jewzaam can I get a review from you here please?

build/deploy.sh
```diff
@@ -10,4 +10,4 @@ fi
 
 # Build the image
 
-make IMAGE_REPOSITORY=$IMAGE_REPOSITORY build
+make IMAGE_REPOSITORY=$IMAGE_REPOSITORY build skopeo-push
```
jewzaam: The `skopeo-push` target in `standard.mk` and `create_catalog_image` in `functions.mk` both are doing `skopeo copy`, I presume in the same way.  The `create_catalog_image` define is called from target `build-catalog-image` in `standard.mk`.

This is just to say we have multiple ways of doing the same thing defined.  Is there anything that will call target `build-catalog-image`?  If not then it should be removed.  If so, we have a risk of having these commands diverging.

lisa:
```suggestion
make IMAGE_REPOSITORY=$IMAGE_REPOSITORY build skopeo-push build-catalog-image
```

This also needs to call `build-catalog-image`

@jewzaam The `skopeo-push` target is only for the operator image, not the OLM Catalog. All of the OLM Catalog build/push is taken care of from the `build-catalog-image` make target (`create_catalog_image` in `functions.mk`).

jharrington22:
@lisa @jewzaam updated, please re-review.

PR #24, opened 2019-05-21
Updated placeholder guage's label

jewzaam:
I know it will go away, but it wasn't consistent w/ other operator placeholder metrics.


jewzaam:
Placeholder metric is gone in master. Closing.

PR #40, opened 2019-08-09
do not filter alerts based on namespaces regex

RiRa12621:
This is commenting out the regex that is filtering alerts out, so that only matching namespaces go to Pagerduty.
The prometheusrules used in the monitoring-operator do not tag alerts in that way, in return no alerts coming from those rules currently go to PagerDuty.
In order to enable alerting, this is commented out, once alertrules in the monitoring operator have more labels, including `namespace` that we can match on, this can be reenabled.

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -265,9 +265,10 @@ func addPDSecretToAlertManagerConfig(r *ReconcileSecret, request *reconcile.Requ
 			"alertname",
 			"severity",
 		},
-		MatchRE: map[string]string{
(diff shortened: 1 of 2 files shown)


RiRa12621:
filtering happens on the cluster-monitoring-operator already.
only ServiceMonitors from namespaces labeled with `openshift.io/cluster-monitoring` are being picked up.
as long as our ServiceMonitors, which customers also cannot deploy due to lack of permissions, are somewhat limited, which they have to be because of the pickup labels, there is almost no way we get metrics from customer namespaces

RiRa12621:
Note that this PR has become more important with the first paying customer on v4

PR #41, opened 2019-08-15
added link2 that will point to 'annotations.runbook'

mwoodson:
https://jira.coreos.com/browse/SREP-1942

OSDv4: configure-alertmanager operator looks for "links" in the "runbook" annotation


pkg/controller/secret/secret_controller_test.go
```diff
@@ -174,6 +174,7 @@ func Test_addPDSecretToAlertManagerConfig(t *testing.T) {
 		Description:    `{{ .CommonLabels.alertname }} {{ .CommonLabels.severity | toUpper }} ({{ len .Alerts }})`,
 		Details: map[string]string{
 			"link":         `{{ if .CommonAnnotations.link }}{{ .CommonAnnotations.link }}{{ else }}https://github.com/openshift/ops-sop/tree/master/v4/alerts/{{ .CommonLabels.alertname }}.md{{ end }}`,
+			"link2":        `{{ if .CommonAnnotations.runbook }}{{ .CommonAnnotations.runbook }}{{ else }}{{ end }}`,
```
jewzaam: Is the config happy with an empty value in `link2`?  Does it become an empty string or just nothing?

jewzaam: I think it should be OK, just yaml at the end, value would be `null`.  I'll approve and we can just confirm in stage with some real alerts.

lisa:
@mwoodson Late to the party, but `runbook_link` is going to be easier to understand than `link2`.

PR #42, opened 2019-08-29
Exclude KubeAPILatencyHigh alert

liamwazherealso:
https://jira.coreos.com/browse/SREP-1922

KubeAPILatencyHigh is setup in prometheusrules to trigger a critical alert at 10 minutes.

Almost every one of these alerts is automatically resolved within 10 minutes of triggering.  SRE has only fixed one and it was a side effect of other issues on a cluster: https://redhat.pagerduty.com/alerts/PK2NDLT


liamwazherealso:
Tested

RiRa12621:
Am I understanding this wrong or is this actually the opposite of what you want to do?
You're adding everything that matches KubeAPILatencyHigh to the route that send alert to pagerduty.
Feel free to correct me if i'm wrong.

fahlmant:
@RiRa12621 This was reverted in #45

PR #46, opened 2019-09-26
Suppress KubeAPILatencyHigh

liamwazherealso:
https://jira.coreos.com/browse/SREP-1922

KubeAPILatencyHigh is setup in prometheusrules to trigger a critical alert at 10 minutes.

Almost every one of these alerts is automatically resolved within 10 minutes of triggering.  SRE has only fixed one and it was a side effect of other issues on a cluster.


liamwazherealso:
@jewzaam https://github.com/openshift/configure-alertmanager-operator/pull/46/files#diff-7782a4838d0c0a8edbf0b736b9b96b61R269 This line modifies the value.

PR #50, opened 2019-10-28
Fix logging and misc

mjudeikis:
Fix logging based on ZAP logger spec.

* usage of the logger is wrong and it generated a lot of noise.
* remove isclean as it makes development hard and does not bring almost any value
* bump sdk to 0.8.2 to align with other operators


jharrington22:
@mjudeikis Looks good could you just squash the non vendor commits?

mjudeikis:
@jharrington22 done

PR #64, opened 2020-02-25
Suppress CPUThrottingHigh alert where container="registry-server"

jewzaam:
https://issues.redhat.com/browse/OSD-2980

Refactored how similar alerts are managed to make it more flexible since this condition requires more than just the `alertname`.

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -24,15 +24,16 @@ import (
 var log = logf.Log.WithName("secret_controller")
 
 var (
-	alertsRouteWarning = []string{
-		"KubeAPILatencyHigh",
+	alertsRouteWarning = []map[string]string{
+		{"alertname": "KubeAPILatencyHigh", "severity": "critical"},
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -491,25 +478,71 @@ func removeConfigFromAlertManager(r *ReconcileSecret, request *reconcile.Request
...
 			&alertmanager.Route{
-				Receiver: "make-it-warning",
+				Receiver: rec,
 				Match: map[string]string{
-					"alertname": alert,
+					"alertname": "KubeAPILatencyHigh",
+					"severity":  "critical",
 				},
 			})
 	}
```
lisa: Why is this block wrapped with `{}`?

jewzaam: To scope the `rec` variable and ensure we don't accidentally use the wrong value for Receiver for Routes we expect to be going to `null`.  Originally was flat and two vars but I could see future revisions with copy and paste landing the wrong value.

PR #66, opened 2020-02-28
Suppress NodeFilesystemSpaceFillingUp{severity="warning"}

jewzaam:
https://issues.redhat.com/browse/OSD-3010

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,14 +1,12 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
-ENV OPERATOR_PATH=/go/src/github.com/openshift/configure-alertmanager-operator
-COPY . ${OPERATOR_PATH}
-WORKDIR ${OPERATOR_PATH}
+COPY . /go/src/github.com/openshift/configure-alertmanager-operator
+WORKDIR /go/src/github.com/openshift/configure-alertmanager-operator
 RUN make gobuild
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
-ENV OPERATOR_PATH=/go/src/github.com/openshift/configure-alertmanager-operator \
-    OPERATOR_BIN=configure-alertmanager-operator
+ENV OPERATOR_BIN=configure-alertmanager-operator
 
 WORKDIR /root/
-COPY --from=builder ${OPERATOR_PATH}/build/_output/bin/${OPERATOR_BIN} /usr/local/bin/${OPERATOR_BIN}
+COPY --from=builder /go/src/github.com/openshift/configure-alertmanager-operator/build/_output/bin/${OPERATOR_BIN} /usr/local/bin/${OPERATOR_BIN}
 LABEL io.openshift.managed.name="configure-alertmanager-operator" \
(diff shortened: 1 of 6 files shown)


jewzaam:
Have reproduced the build issue finally and will poke at this soon.

jewzaam:
@cblecker not sure what is going on with this PR, but by removing `ENV OPERATOR_PATH` and hard coding it works.  The problem is in the `COPY --from` and no combination of ENV and ARG would honor any value for that from location.  Locally I use podman (which is now supported in this PR) and I wonder if something in the prow back-end changed such as swapping from docker to podman.. but that's just speculation.

I'm not super happy hard coding these paths BUT they're not going to change for an operator.  And we were setting them in two places already since each stage in a multi-stage build needed the ENV set.

standard.mk
```diff
@@ -29,35 +29,41 @@ MAINPACKAGE=./cmd/manager
 GOENV=GOOS=linux GOARCH=amd64 CGO_ENABLED=0
 GOFLAGS=-gcflags="all=-trimpath=${GOPATH}" -asmflags="all=-trimpath=${GOPATH}"
 
+CONTAINER_ENGINE=`which podman 2>/dev/null || which docker 2>/dev/null`
```
lisa:
```suggestion
CONTAINER_ENGINE=$(shell which podman 2>/dev/null || which docker 2>/dev/null)
```

build/Dockerfile
```diff
@@ -1,14 +1,12 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
-ENV OPERATOR_PATH=/go/src/github.com/openshift/configure-alertmanager-operator
-COPY . ${OPERATOR_PATH}
-WORKDIR ${OPERATOR_PATH}
+COPY . /go/src/github.com/openshift/configure-alertmanager-operator
+WORKDIR /go/src/github.com/openshift/configure-alertmanager-operator
 RUN make gobuild
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
-ENV OPERATOR_PATH=/go/src/github.com/openshift/configure-alertmanager-operator \
```
lisa: Should this be a build-arg (`ARG`) in order to support referencing it in the Dockerfile? I'm not opposed to the hardcoding of it, but mainly curious.

standard.mk
```diff
@@ -29,35 +29,41 @@ MAINPACKAGE=./cmd/manager
...
+
 TESTTARGETS := $(shell go list -e ./... | egrep -v "/(vendor)/")
 # ex, -v
 TESTOPTS :=
 
 ALLOW_DIRTY_CHECKOUT?=false
 
+.PHONY: testx
+testx:
+	$(shell echo "$(CONTAINER_ENGINE)")
```
lisa:
```suggestion
	@echo $(CONTAINER_ENGINE)
```

This should do the same, without a subshell.

build/Dockerfile
jewzaam: I tried. It didn't work.

standard.mk
```diff
@@ -29,35 +29,41 @@ MAINPACKAGE=./cmd/manager
...
+
```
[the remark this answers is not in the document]
jewzaam: No, removing.

PR #71, opened 2020-03-26
Suppress SLAUptimeSRE alert

superfunhappytime:
This is a paired commit to suppress SRE members being alerted for OSD-3220 (SLA tracking)
https://github.com/openshift/managed-cluster-config/pull/265


jewzaam:
/approve
/lgtm
/hold 

Can remove hold once we have finalized https://github.com/openshift/managed-cluster-config/pull/265

PR #73, opened 2020-04-09
Refactor to allow more flexible routing.

jewzaam:
Specifically to route ES and fluentd alerts to PD.

https://issues.redhat.com/browse/OSD-3326

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -23,7 +23,7 @@ import (
 
 var log = logf.Log.WithName("secret_controller")
 
-var (
+const (
 	secretKeyPD = "PAGERDUTY_KEY"
 
 	secretKeyDMS = "SNITCH_URL"
@@ -33,7 +33,35 @@ var (
 	secretNameDMS = "dms-secret"
 
 	secretNameAlertmanager = "alertmanager-main"
+
+	// anything routed to "null" receiver does not get routed to PD
+	receiverNull = "null"
+
+	// anything routed to "make-it-warning" receiver has severity=warning
+	receiverMakeItWarning = "make-it-warning"
+
+	// anything routed to "pagerduty" will alert/notify SREP
+	receiverPagerduty = "pagerduty"
+
+	// anything going to Dead Man's Snitch (watchdog)
+	receiverWatchdog = "watchdog"
+
+	// the default receiver used by the route used for pagerduty
+	defaultReceiver = receiverNull
+
+	// global config for PagerdutyURL
+	pagerdutyURL = "https://events.pagerduty.com/v2/enqueue"
 )
+
+var _ reconcile.Reconciler = &ReconcileSecret{}
+
+// ReconcileSecret reconciles a Secret object
+type ReconcileSecret struct {
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	client client.Client
+	scheme *runtime.Scheme
+}
 
 // Add creates a new Secret Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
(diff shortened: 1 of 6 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -89,6 +73,184 @@ type ReconcileSecret struct {
...
+
+		// general: route anything in core namespaces to PD
+		{Receiver: receiverPagerduty, MatchRE: map[string]string{"namespace": alertmanager.PDRegex}},
+		// fluentd: route any fluentd alert to PD
+		// https://issues.redhat.com/browse/OSD-3326
+		{Receiver: receiverPagerduty, Match: map[string]string{"job": "fluentd"}},
+		{Receiver: receiverPagerduty, Match: map[string]string{"alertname": "FluentdNodeDown"}},
+		// elasticsearch: route any ES alert to PD
+		// https://issues.redhat.com/browse/OSD-3326
+		{Receiver: receiverPagerduty, Match: map[string]string{"cluster": "elasticsearch"}},
```
rogbas, line 133: I'm not familiar with this `cluster` parameter from the alert being used like that. Are we sure that's how we filter for elasticsearch related alerts?

pkg/controller/secret/secret_controller.go
```diff
@@ -89,6 +73,184 @@ type ReconcileSecret struct {
...
+		Global: &alertmanager.GlobalConfig{
+			ResolveTimeout: "5m",
+			PagerdutyURL:   pagerdutyURL,
+		},
+		Route: &alertmanager.Route{
+			Receiver: defaultReceiver,
+			GroupByStr: []string{
+				"job",
+			},
+			GroupWait:      "30s",
```
rogbas, line 242: Shouldn't we keep those parameters as global constants?

pkg/controller/secret/secret_controller.go
```diff
@@ -218,16 +356,15 @@ func getSecretKey(r *ReconcileSecret, request *reconcile.Request, secretname str
 // This is equivalent to `oc get secrets -n openshift-monitoring alertmanager-main`.
 // It specifically extracts the .data "alertmanager.yaml" field, and loads it into a resource
 // of type Config, enabling it to be marshalled and unmarshalled as needed.
-func getAlertManagerConfig(r *ReconcileSecret, request *reconcile.Request) alertmanager.Config {
```
rogbas, line 221: After this renaming this function the comment above is still referring to the previous name.

pkg/controller/secret/secret_controller.go
jewzaam: Need something to tag on.  All the alerts coming from elasticsearch include the label `cluster` with a value of `elasticsearch` so I picked that.  I think other common labels.  Ideally we'd get a `namespace` label and be done but we don't get it.  If this is a problem I can look again at the alerts.

pkg/controller/secret/secret_controller.go
jewzaam: We could but didn't because it isn't something I expect we'll change.  The old controller code didn't manage the global config at all.  It merged the bits we wanted into the default provided by OCP.  This PR change things to take 100% control of the alertmanager-main secret.

pkg/controller/secret/secret_controller.go
rogbas: If this label is assigned to all `elasticsearch` alerts +1

pkg/controller/secret/secret_controller.go
```diff
@@ -362,168 +380,16 @@ func updateAlertManagerConfig(r *ReconcileSecret, request *reconcile.Request, am
...
-			amconfig.Route.Routes = removeFromRoutes(amconfig.Route.Routes, i)
-		}
-	}
-
-	if nullrouteabsent {
-		nullreceiver := &alertmanager.Receiver{Name: "null"}
-		amconfig.Receivers = append(amconfig.Receivers, nullreceiver)
-	}
-
-	// If watchdog is being removed, put the system default route and receiver back into place.
```
cblecker, line 497: Has this been tested if DMS is removed? Does the default receiver go back in?
(this would be the case where someone adds `noalerts` to a cluster and the syncset removes the secret)

jewzaam: The "null" route is now always set, it wasn't before.  This no longer tries to restore to system defaults.  It just has a bare bones route to "null" receiver.

jewzaam:
@cblecker @rogbas fixed a few things:
- "null" receiver is always present
- alert names and expr fixed

Deployed in stage and verified.

PR #76, opened 2020-04-27
Use Go 1.13 & Go modules; update operator-sdk; update prometheus/client_golang

grdryn:
We're planning to add a CRD for layered product alert configuration,
and the operator-sdk will be used to generate that CRD. A lot has
changed since the operator-sdk that was previously specified here
(0.8.2), including CRD and code generation; so I think it makes sense
to update it before attempting that work.

The newer versions of the operator-sdk CLI expect a Go modules
project, rather than a dep-based project; and Go 1.13+. I chose to
stop at 0.16.0 for now, since 0.17.0 bumps the Kubernetes dependency
versions to v1.17.4, which is past the current 1.16.z of Kubernetes in
OpenShift 4.3.

Updating to Go modules defaulted to a newer version of the
prometheus/client_golang library which had removed some stuff here
that was deprecated. I initially added a replace to the go.mod to work
around that, but it turns out that it was a very easy fix so I've
added that here too.

Since `go mod vendor` only stores packages that are referenced, a lot
of files have been removed from the vendor directory here.

I put some details into each commit message highlighting the steps I
took at each stage, in the hopes that it might be useful to verify I haven't
done anything nasty. I think the following will also help with that:

```bash
git checkout gomodules-1.13-operator-sdk
rm -rf vendor
go mod tidy
go mod vendor

git status
# there shouldn't be any changed files
```

I'm not sure what the best way to try this out is. What I did (and
worked fine) was to update the CSV for this operator in an OSD dev
cluster.


jewzaam:
looks reasonable, will look for someone else to +1 so we have more eyes on it

/approve

cblecker:
/hold
This is good, but needs close inspection, and some matching changes in openshift/release.

standard.mk
```diff
@@ -27,7 +27,7 @@ OPERATOR_DOCKERFILE?=build/Dockerfile
 BINFILE=build/_output/bin/${OPERATOR_NAME}
 MAINPACKAGE=./cmd/manager
 GOENV=GOOS=linux GOARCH=amd64 CGO_ENABLED=0
-GOFLAGS=-gcflags="all=-trimpath=${GOPATH}" -asmflags="all=-trimpath=${GOPATH}"
+GOFLAGS=-gcflags="all=-trimpath=${GOPATH}" -asmflags="all=-trimpath=${GOPATH}" -mod=vendor
```
cblecker, line 30: We should ditch vendor and use modules exclusively.

grdryn:
Thanks for the review folks!

@cblecker I'm not sure exactly what changes are needed in openshift/release (and for which branches, e.g. should I update just master or also 4.6?), but here's maybe a start: https://github.com/openshift/release/pull/8962

Wrt ditching vendoring, I'm all for that :) The base builder image currently sets `GOFLAGS` in a way that expects vendoring by default:

https://github.com/openshift/release/blob/master/projects/origin-release/golang-1.13/Dockerfile#L14

There's a PR to add a Go 1.14 builder image there too, where I've suggested maybe dropping that defaulted `-mod=vendor` option: https://github.com/openshift/release/pull/8542/files#r422085608

For now, I'll push a new commit to this PR to remove the vendor directory, and to unset that flag in our builder image stage in the Dockerfile.

jewzaam:
@cblecker looks like the changes you've requested are done.  ptal when you can

grdryn:
Rebased on latest master.

jharrington22:
@cblecker did we verify with the operator framework team that this is ok to do? Last time I checked with them while its possible to upgrade they did not recommend going past 0.8.2.

grdryn:
@jharrington22 that's interesting, do you recall why that recommendation was made? 0.8.2 is very old at this stage. Do you use this operator on older OpenShift versions or something?

cblecker:
There some things we should do to standardize the Makefile, and update the docker image to use UBI.. but those can be done in a follow up.

/lgtm
/approve
/hold cancel

cblecker:
@jharrington22 we have a number of operators on osk 0.16

jharrington22:
> @jharrington22 we have a number of operators on osk 0.16

Yes I know we have operators running at the version but the SDK team was concerned about upgrading from < 0.8.2. I guess we'll see how this goes!

PR #83, opened 2020-05-25
Mute all alerts to sre-p from openshift-operators namespace

oddtazz:
Mute all alerts to sre-p from openshift-operators namespace

https://issues.redhat.com/browse/OSD-3569

https://redhat.pagerduty.com/alerts/P1Q5QBZ

This is a namespace where customers can deploy operators. SRE doesn't support those operators. Therefore, SRE shouldn't be alerted if there are any problems in that namespace.

Done:

    SREP is not alerted for any issues in openshift-operators namespace

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -115,6 +115,8 @@ func createPagerdutyRoute() *alertmanager.Route {
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "NodeFilesystemSpaceFillingUp", "severity": "warning"}},
 		// https://issues.redhat.com/browse/OSD-2611
 		{Receiver: receiverNull, Match: map[string]string{"namespace": "openshift-customer-monitoring"}},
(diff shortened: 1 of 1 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -115,6 +115,8 @@ func createPagerdutyRoute() *alertmanager.Route {
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "NodeFilesystemSpaceFillingUp", "severity": "warning"}},
 		// https://issues.redhat.com/browse/OSD-2611
 		{Receiver: receiverNull, Match: map[string]string{"namespace": "openshift-customer-monitoring"}},
+                // https://issues.redhat.com/browse/OSD-3569
+                {Receiver: receiverNull, Match: map[string]string{"namespace": "openshift-operators"}},
```
cblecker, line 119: Looks like whitespace is off?

oddtazz: Fixed :)

PR #84, opened 2020-05-25
Add codecov configuration

cblecker:


2uasimojo:
Should this/a codecov config be part of the [SOP](https://github.com/openshift/ops-sop/blob/master/services/codecov.md)?

2uasimojo:
> Should this/a codecov config be part of the [SOP](https://github.com/openshift/ops-sop/blob/master/services/codecov.md)?

Done via https://github.com/openshift/ops-sop/pull/799

PR #92, opened 2020-07-08
Silence TargetDown for redhat- NS

dofinn:


jewzaam:
/lgtm
/approve

context: https://github.com/openshift/cluster-monitoring-operator/pull/835/files

RiRa12621:
Are we sure RHMI also doesn't care about those alerts? if they do, they should get it, if they don't, everything in their namespace should go /dev/null

jewzaam:
https://issues.redhat.com/browse/OSD-4373

PR #94, opened 2020-08-17
redirect the warnings for etcd to null

bmeng:
https://issues.redhat.com/browse/OSD-4631

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -125,6 +125,8 @@ func createPagerdutyRoute() *alertmanager.Route {
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "ImagePruningDisabled"}},
 		// https://issues.redhat.com/browse/OSD-3794
 		{Receiver: receiverNull, Match: map[string]string{"severity": "info"}},
+		// https://issues.redhat.com/browse/OSD-4631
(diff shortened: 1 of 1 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -125,6 +125,8 @@ func createPagerdutyRoute() *alertmanager.Route {
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "ImagePruningDisabled"}},
 		// https://issues.redhat.com/browse/OSD-3794
 		{Receiver: receiverNull, Match: map[string]string{"severity": "info"}},
+		// https://issues.redhat.com/browse/OSD-4631
+		{Receiver: receiverNull, MatchRE: map[string]string{"alertname": "^etcd.*"}, Match: map[string]string{"namespace": "openshift-monitoring", "severity": "warning"}},
```
jewzaam, line 129:
```suggestion
		{Receiver: receiverNull, MatchRE: map[string]string{"alertname": "^etcd.*"}, Match: map[string]string{"namespace": "openshift-etcd", "severity": "warning"}},
```

Taking a look at a subset of alerts in prod the label `namespace` has a value of `openshift-etcd` always.  Alternately the criteria on `namespace` could be removed, leaving only `severity`.

PR #98, opened 2020-09-07
Add ClusterID to alert message

NautiluX:
* Simplify login flow from alert to cluster
* ClusterID is read from ClusterVersion

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -425,7 +425,10 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 	}
 
 	// create the desired alertmanager Config
-	clusterID := r.getClusterID()
+	clusterID, err := r.getClusterID()
+	if err != nil {
+		log.Error(err, "Error reading cluster id.")
+	}
 	alertmanagerconfig := createAlertManagerConfig(pagerdutyRoutingKey, watchdogURL, consoleUrl, clusterID)
 
 	// write the alertmanager Config
@@ -436,14 +439,13 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 	return reconcile.Result{}, nil
 }
 
-func (r *ReconcileSecret) getClusterID() string {
+func (r *ReconcileSecret) getClusterID() (string, error) {
 	var version configv1.ClusterVersion
 	err := r.client.Get(context.TODO(), client.ObjectKey{Name: "version"}, &version)
 	if err != nil {
-		log.Error(err, "Error reading cluster id.")
-		return ""
+		return "", err
 	}
-	return string(version.Spec.ClusterID)
+	return string(version.Spec.ClusterID), nil
 }
 
 // secretInList takes the name of Secret, and a list of Secrets, and returns a Bool
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -21,6 +21,7 @@ import (
 	"github.com/openshift/configure-alertmanager-operator/pkg/metrics"
 	alertmanager "github.com/openshift/configure-alertmanager-operator/pkg/types"
 
+	configv1 "github.com/openshift/api/config/v1"
 	yaml "gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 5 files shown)


RiRa12621:
/lgtm
Maybe the unrelated logging improvements should have been a separate commit tho

jewzaam:
Similar to https://github.com/openshift/configure-alertmanager-operator/pull/99 and expect some conflicts, please work with @cblecker on how you want to merge these.

cblecker:
@NautiluX #99 has merged. You're clear to rebase :)

NautiluX:
rebased to master

NautiluX:
@jewzaam I think we can cancel the hold now if you agree.

pkg/controller/secret/secret_controller.go
```diff
@@ -433,6 +436,16 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileSecret) getClusterID() string {
+	var version configv1.ClusterVersion
+	err := r.client.Get(context.TODO(), client.ObjectKey{Name: "version"}, &version)
+	if err != nil {
+		log.Error(err, "Error reading cluster id.")
```
cblecker, line 443: Can we return this as an `error`, and then log it in the main loop (with the log with variables)?

pkg/controller/secret/secret_controller_test.go
```diff
@@ -297,13 +299,13 @@ func Test_createPagerdutyRoute(t *testing.T) {
 }
 
 func Test_createPagerdutyReceivers_WithoutKey(t *testing.T) {
-	assertEquals(t, 0, len(createPagerdutyReceivers("", "")), "Number of Receivers")
+	assertEquals(t, 0, len(createPagerdutyReceivers("", "", "fake-cluster-id")), "Number of Receivers")
```
cblecker, line 302: In this instance, all three strings should be empty

pkg/controller/secret/secret_controller_test.go
```diff
@@ -297,13 +299,13 @@ func Test_createPagerdutyRoute(t *testing.T) {
...
 }
 
 func Test_createPagerdutyReceivers_WithKey(t *testing.T) {
 	key := "abcdefg1234567890"
 
-	receivers := createPagerdutyReceivers(key, exampleConsoleUrl)
+	receivers := createPagerdutyReceivers(key, exampleConsoleUrl, "fake-cluster-id")
```
cblecker, line 308: Can we make `"fake-cluster-id"` a const, and then use it in all these places? This just ensures that the test and the example string always match

pkg/controller/secret/secret_controller_test.go
```diff
@@ -630,7 +651,7 @@ func Test_ReconcileSecrets(t *testing.T) {
 
 		// Create the secrets for this specific test.
 		if tt.amExists {
-			writeAlertManagerConfig(reconciler, createAlertManagerConfig("", "", ""))
+			writeAlertManagerConfig(reconciler, createAlertManagerConfig("", "", "", "fake-cluster-id"))
```
cblecker, line 654: This should also be a blank string (the test confirms that the blank string is overwritten)

pkg/controller/secret/secret_controller_test.go
```diff
@@ -644,10 +665,11 @@ func Test_ReconcileSecrets(t *testing.T) {
 			createSecret(reconciler, secretNamePD, secretKeyPD, pdKey)
 		}
 
-		configExpected := createAlertManagerConfig(pdKey, wdURL, exampleConsoleUrl)
+		configExpected := createAlertManagerConfig(pdKey, wdURL, exampleConsoleUrl, "fake-cluster-id")
 
 		verifyInhibitRules(t, configExpected.InhibitRules)
 
+		createClusterVersion(reconciler)
```
cblecker, line 672: I think this needs to happen sooner, doesn't it? Like, line 647?

pkg/controller/secret/secret_controller.go
NautiluX: Sure, I don't have any preference about where we hide the error ;)

pkg/controller/secret/secret_controller_test.go
NautiluX: done

pkg/controller/secret/secret_controller_test.go
NautiluX: make sense. Done

pkg/controller/secret/secret_controller_test.go
NautiluX: yes, now that the test actually does something I need to move it :)

NautiluX:
Thanks for your review @cblecker, addressed all your comments in the latest commit.

PR #99, opened 2020-09-07
Provide console URL into PD alerts

cblecker:
This grabs the console URL from the system configmap (same thing `oc` does here: https://github.com/openshift/oc/blob/f2a4a0375cc7b0eacb5467a0214303983e1151d6/pkg/cli/whoami/whoami.go#L115-L130) and then puts it into the alert configuration.

This will speed up alert response time, as an SRE can click immediately from the alert to access the cluster's console.

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -17,6 +17,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/openshift/configure-alertmanager-operator/config"
 	"github.com/openshift/configure-alertmanager-operator/pkg/metrics"
 	alertmanager "github.com/openshift/configure-alertmanager-operator/pkg/types"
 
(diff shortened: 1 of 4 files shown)


cblecker:
/hold
Holding for discussion

cblecker:
Reviewing the test coverage report, the misses are just negative error checking. This should be safe to merge.

cblecker:
Found an issue in testing:
Because of the way the caching client was set up, it wasn't able to grab the configmap from `openshift-config-managed`. Made some changes to allow it to read this. I don't know how I feel about them though

cblecker:
Example of how this looks in the PD UI:
example PD alert UI

pkg/controller/secret/secret_controller.go
```diff
@@ -379,8 +404,14 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 		watchdogURL = readSecretKey(r, &request, secretNameDMS, secretKeyDMS)
 	}
 
+	// grab the console URL for PD alerts
+	consoleUrl, err := r.getWebConsoleUrl()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
```
jewzaam, line 411: If we're not able to get the console URL it shouldn't block configuration of alertmanager.  We can still get stuff done, we just are not quite as efficient.  But not having alerts configured is a big problem.  Recommend we simply ignore the error and set consoleUrl to something that won't cause the config to blow up.

jewzaam (changes requested):

jewzaam:
Similar to https://github.com/openshift/configure-alertmanager-operator/pull/98 and expect some conflicts, please work with @NautiluX on how you want to merge these.

NautiluX:
> Similar to #98 and expect some conflicts, please work with @NautiluX on how you want to merge these.

As mentioned on Slack I think both PRs provide value. Happy to rebase mine after this is merged.

cblecker:
@jewzaam Addressed comments

NautiluX (commented):
/lgtm

PR #101, opened 2020-09-08
Boilerplate bootstrap

2uasimojo:
Initial setup of [boilerplate](https://github.com/openshift/boilerplate/) with the [openshift/golang-osd-operator](https://github.com/openshift/boilerplate/tree/7ec393872cc24919a85508622e547fc4d1e0097f/boilerplate/openshift/golang-osd-operator) convention.

Things of note:
- project.mk goes away, replaced by its counterpart in the boilerplate convention.
- Almost all of standard.mk is subsumed by its counterpart in the boilerplate convention. The remainder is retained for the time being, but moved into the main Makefile.
- functions.mk is also retained for now. A future update to boilerplate will standardize this logic (and the remaining bespoke bits from standard.mk).
- The `gocheck` target brought in by boilerplate uses golangci-lint instead of gofmt + go vet. The failures are resolved here. Most of the fixes are legit, but a couple are spoofed, with TODOs added, to be addressed later.


2uasimojo:
This is failing because the golang-1.13 container doesn't come with the PyYAML module by default. Eventually we'll standardize on a more complete container for boilerplate consumers to use, but for now, https://github.com/openshift/boilerplate/pull/40 ensures the module is installed "by hand". Once that lands, I'll update the boilerplate here and we should be green (or at least get to the _next_ failure :P).

2uasimojo:
Editorial: This is shaping up to be a good case study for
- Bring in boilerplate
- Remove stuff brought in by boilerplate (`make` targets and whatnot)
- Fix stuff exposed by boilerplate (in this case due to cutover to golangci-lint)
- Observe a breakage that needs to be fixed "upstream" (in the boilerplate repo)
- Fix it
- Update again to bring in that fix

2uasimojo:
So we still have a problem with PyYAML. I believe the fix will be to ensure that python3-devel is installed in the container.

What is really baking my noodle, though, is that after the first failure I went in and spun up a container using the actual image from the CI environment, and installed PyYAML using the [exact command added via openshift/boilerplate#40](https://github.com/openshift/boilerplate/pull/40/files#diff-6efb8dd653c9062efcfb26adab7756fcR125) and it worked fine.

A discrepancy between container-in-CI and same-container-by-hand also showed up in https://github.com/openshift/boilerplate/pull/33. I worked around it there; but clearly we're going to have to nail it down.

@stevekuznetsov :arrow_up:
Unless you have ideas off the top, I guess I'll ping @dptp-helpdesk tomorrow.

2uasimojo:
/test test

We're meeting in a few minutes to go over this. Getting a fresh result to use.

2uasimojo:
For those of you following along at home, the discrepancy in the container environments is: When I run via docker, I'm root in the container. In the CI environment, I'm not. All else flows from that.

So having figured out how to reproduce correctly in the CI env, we have a higher degree of confidence that https://github.com/openshift/boilerplate/pull/42 will get the yaml check passing. Will wait for that to merge, and then respin the third commit in this PR.

jewzaam:
/lgtm
/approve
/hold

@2uasimojo remove hold when ready to merge, looks good to me.

2uasimojo:
Thanks @jewzaam, let's do this!

/hold cancel

PR #104, opened 2020-09-14
Assign readonly Clusterrole to serviceaccount

NautiluX:


manifests/03_role_binding.yaml
```diff
@@ -25,3 +25,16 @@ subjects:
 - kind: ServiceAccount
   name: configure-alertmanager-operator
   namespace: openshift-monitoring
+---
+apiVersion: authorization.openshift.io/v1
+kind: ClusterRoleBinding
```
jewzaam, line 30: Given this is granted at a cluster scope we should remove "secrets" and "configmaps" from clusterrole "configure-alertmanager-operator-view".

jewzaam (changes requested):

manifests/03_role_binding.yaml
NautiluX: agree, since the role was never bound to the user we should be able to remove those.

PR #109, opened 2020-10-09
Fix skopeo push for container

cblecker:
Replaces the `skopeo-push` step required by app-sre CI. This was removed as a part of #101 / 69205363e29595168faa965cee0d226b45afd344

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,17 @@ include functions.mk
 
 CATALOG_REGISTRY_ORGANIZATION?=app-sre
 
+.PHONY: skopeo-push
+skopeo-push: build
+	skopeo copy \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -16,6 +16,17 @@ include functions.mk
 
 CATALOG_REGISTRY_ORGANIZATION?=app-sre
 
+.PHONY: skopeo-push
+skopeo-push: build
```
2uasimojo, line 20: `build` is redundant (included in the call from `deploy.sh` but meh.

2uasimojo (commented):
/lgtm

I'll investigate the upstream fix next week.

PR #110, opened 2020-10-15
use mirrored yq image

jewzaam:


lisa:
@jewzaam Why did go.mod and go.sum change for a Makefile change?

lisa (approved):
/lgtm
/approve

PR #113, opened 2020-10-19
Swap the build-catalog-image variables back in

cblecker:
This makes it so that these bits can be copied verbatim into other repos. I've tested this locally and it does the trick.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,5 +29,5 @@ skopeo-push: build
 
 .PHONY: build-catalog-image
 build-catalog-image:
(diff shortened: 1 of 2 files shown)


cblecker:
mine outputs to this:
```
++ curl -s https://gitlab.cee.redhat.com/service/app-interface/raw/master/data/services/osd-operators/cicd/saas/saas-configure-alertmanager-operator.yaml
++ /usr/local/bin/docker run --rm -i quay.io/app-sre/yq yq r - 'resourceTemplates[*].targets(namespace.ef==/services/osd-operators/namespaces/hivep01ue1/cluster-scope.yml).ref'
```

jharrington22 (approved):
/lgtm

PR #125, opened 2020-11-18
re-route the flipping alerts HAProxyReloadFail

bmeng:
The alert HAProxyReloadFail is annoying, and we can do nothing except wait for it auto resolving.

We got more than 160 alerts for it for the past 2 weeks, and none of them is actionable.

Temporarily lower the severity till we got the fix landed.

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -147,6 +147,8 @@ func createPagerdutyRoute() *alertmanager.Route {
 		{Receiver: receiverNull, MatchRE: map[string]string{"namespace": alertmanager.PDRegexLP}, Match: map[string]string{"alertname": "TargetDown"}},
 		// https://issues.redhat.com/browse/OSD-5544
 		{Receiver: receiverNull, MatchRE: map[string]string{"job_name": "^elasticsearch.*"}, Match: map[string]string{"alertname": "KubeJobFailed", "namespace": "openshift-logging"}},
+		// Suppress the alerts and use HAProxyReloadFailSRE instead (openshift/managed-cluster-config#600)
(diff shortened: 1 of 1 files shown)


bmeng:
@jewzaam Can we make this till the bug fixed?

jewzaam:
I'm OK with this but there's no easy way to apply only to < 4.5.17 so I expect we'll carry this until someone says "why are these low and not high" in PD.  So there's risk.

Could we carry an SRE version of the alert that sets `for` long enough it doesn't fire for the noise we have seen in PD?  In that case we route the current alert to null receiver, create a new alert, and include in observability processes.

@RiRa12621 thoughts?

RiRa12621:
+1 for the custom version.
Since the metric itself changed in the fix from counter to gauge, our adjusted version will then never force
I think this is a case that you mention in the mail @jewzaam : we need to have a bit of automation to look after temp changes that relate to BZs.
I'll come up with a plan tomorrow

pkg/controller/secret/secret_controller.go
```diff
@@ -150,6 +150,8 @@ func createPagerdutyRoute() *alertmanager.Route {
 
 		// https://issues.redhat.com/browse/OSD-1922
 		{Receiver: receiverMakeItWarning, Match: map[string]string{"alertname": "KubeAPILatencyHigh", "severity": "critical"}},
+		// Lower the alerts till we upgrade to version later than 4.5.17 https://bugzilla.redhat.com/show_bug.cgi?id=1885688
+		{Receiver: receiverMakeItWarning, Match: map[string]string{"alertname": "HAProxyReloadFail", "severity": "critical"}},
```
dofinn, line 154: @bmeng could you please update this to receiverNull ? we can then commit this with custom SRE alert.

bmeng: Updated.

cblecker (approved):

PR #126, opened 2020-11-30
Add to node alert inhibition rules

cblecker:
Add a few things to the node alert inhibition rules:
- If a node is `KubeNodeUnreachable`, then we can inhibit alerting on `KubeNodeNotReady`
- If `KubeNodeNotReady` is firing, then we can inhibit the following alerts:
  - `KubeDaemonSetMisScheduled`
  - `SDNPodNotReady`
  - `TargetDown`

### DIFF
--- a/pkg/controller/secret/secret_controller_test.go
+++ b/pkg/controller/secret/secret_controller_test.go
@@ -248,13 +248,61 @@ func verifyInhibitRules(t *testing.T, inhibitRules []*alertmanager.InhibitRule) 
 		},
 		{
 			SourceMatch: map[string]string{
+				"alertname": "KubeNodeUnreachable",
+			},
+			TargetMatchRE: map[string]string{
+				"alertname": "KubeNodeNotReady",
+			},
+			Equal: []string{
+				"node",
+				"instance",
+			},
+			Expected: true,
+		},
+		{
+			SourceMatch: map[string]string{
 				"alertname": "KubeNodeNotReady",
 			},
 			TargetMatchRE: map[string]string{
 				"alertname": "KubeDaemonSetRolloutStuck",
 			},
 			Equal: []string{
-				"namespace",
+				"instance",
+			},
+			Expected: true,
+		},
+		{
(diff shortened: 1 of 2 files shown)


pkg/controller/secret/secret_controller_test.go
```diff
@@ -254,7 +267,42 @@ func verifyInhibitRules(t *testing.T, inhibitRules []*alertmanager.InhibitRule)
...
+				"instance",
+			},
+			Expected: true,
+		},
+		{
+			SourceMatch: map[string]string{
+				"alertname": "KubeNodeNotReady",
+			},
+			TargetMatchRE: map[string]string{
+				"alertname": "TargetDown",
```
RiRa12621, line 303: Should that really fire if we drop a node?

cblecker: It does, yes. TargetDown is if a service is missing pods. If a node is down, then the pods will be removed from the service.

RiRa12621: Isn't that only for single service pods or all pods of a service on a single node?

cblecker: Sorry, I don't quite understand your question.

RiRa12621: Sorry for not being clear.
A target should normally be a service and that service should have multiple backends, namely pods.
So the alert for target down should only fire if all pods of a given service are not reacting to the prometheus request.
That can technically only happen when a node is unresponsive if all pods of one service are on the same node or that service only has one pod attached.

cblecker: The expression for TargetDown is:
```
100 * (count by(job, namespace, service) (up == 0) / count by(job, namespace, service) (up)) > 10
```

So if more than 10% of the pods in a service are down, then it will fire.

RiRa12621: thanks for looking that up.

PR #131, opened 2020-12-23
inhibit ElasticsearchClusterNotHealthy if its csv is not in success

yithian:


yithian:
depends on https://github.com/openshift/elasticsearch-operator/pull/609

yithian:
/hold

until after change freeze

pkg/controller/secret/secret_controller.go
```diff
@@ -407,6 +407,17 @@ func createAlertManagerConfig(pagerdutyRoutingKey, watchdogURL, consoleUrl strin
 					"alertname": "KubeDeploymentReplicasMismatch",
 				},
 			},
+			{
+				Equal: []string{
+					"dummylabel",
```
yithian, line 412: NB this label obviously won't match and that's both ok and expected. When a label is missing (or empty) on both source and target, the rule will apply (see: [docs](https://www.prometheus.io/docs/alerting/latest/configuration/#inhibit_rule) ).

If there wasn't a label here, the tests exploded spectacularly, so I figured a label that would never match is the next best thing.

cblecker: Is it possible to get a code comment describing this? I figure this is something we will be scratching our heads about in the future

yithian:
/unhold

this should be fine to land ahead of elasticsearch-operator 4.7

PR #132, opened 2020-12-26
Update inhibit rules

cblecker:
- Adds new silence for `ClusterOperatorDegraded` with the `IdentityProviderConfig_Error` condition ([OSD-6215](https://issues.redhat.com/browse/OSD-6215))
- Collapses inhibition rules into smaller, more readable rules
- Updates tests accordingly

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -145,6 +145,8 @@ func createPagerdutyRoute() *alertmanager.Route {
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "HAProxyReloadFail", "severity": "critical"}},
 		// https://issues.redhat.com/browse/OHSS-2163
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "PrometheusRuleFailures"}},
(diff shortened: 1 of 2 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -145,6 +145,8 @@ func createPagerdutyRoute() *alertmanager.Route {
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "HAProxyReloadFail", "severity": "critical"}},
 		// https://issues.redhat.com/browse/OHSS-2163
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "PrometheusRuleFailures"}},
+		// https://issues.redhat.com/browse/OSD-6215
```
RiRa12621, line 148: Have we attempted a change upstream?

cblecker: No, and I don't think this makes sense for an upstream change.

Reason: In OCP, a misconfigured IDP can be a big deal. The end user should know and understand that they won't be able to log into their own cluster. In OSD, this alert is something the customer has the ability to fix in OCM, but is completely unactionable by SRE.

PR #134, opened 2021-01-15
Further node replace inhibitions

sedroche:
Jira: https://issues.redhat.com/browse/OSD-5688 (Adjust alerts triggered when Machine health check replaces a node)
This is building on https://github.com/openshift/configure-alertmanager-operator/pull/126

Changes:
- From https://github.com/openshift/configure-alertmanager-operator/pull/126 i'm changing `TargetDown` and `SDNPodNotReady` to be inhibited if `KubeNodeUnreachable`  instead of `KubeNodeNotReady` because both those fire after 10m and `KubeNodeNotReady` is 15m so there is a 5 min gap where those alerts reach PD
- From https://github.com/openshift/configure-alertmanager-operator/pull/126 Removing `instance` match label on `TargetDown` as `TargetDown` doesn't have `instance`
-  Then just inhibiting `KubeDeploymentReplicasMismatch`, `KubeStatefulSetReplicasMismatch`, `KubePodNotReady` behind  `KubeNodeNotReady`

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -341,6 +341,15 @@ func createAlertManagerConfig(pagerdutyRoutingKey, watchdogURL, consoleUrl strin
 				},
 			},
 			{
+				// node being Unreachable may also trigger certain pods being unavailable
+				SourceMatch: map[string]string{
+					"alertname": "KubeNodeUnreachable",
+				},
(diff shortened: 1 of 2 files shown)


cblecker:
@sedroche Can you please rebase your changes? Thanks!

sedroche:
@cblecker Rebased :+1:

PR #135, opened 2021-01-20
Add label to Pagerduty route

sam-nguyen7:
Per [OSD-5872](https://issues.redhat.com/browse/OSD-5872) this adds the label "prometheus" with the value of "openshfit-monitoring/k8s" to the PagerDuty route.

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -152,16 +152,17 @@ func createPagerdutyRoute() *alertmanager.Route {
 		{Receiver: receiverMakeItWarning, Match: map[string]string{"alertname": "KubeAPILatencyHigh", "severity": "critical"}},
 
 		// https://issues.redhat.com/browse/OSD-3086
-		{Receiver: receiverPagerduty, MatchRE: map[string]string{"exported_namespace": alertmanager.PDRegex}},
+		// https://issues.redhat.com/browse/OSD-5872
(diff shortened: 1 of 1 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -152,7 +152,8 @@ func createPagerdutyRoute() *alertmanager.Route {
 		{Receiver: receiverMakeItWarning, Match: map[string]string{"alertname": "KubeAPILatencyHigh", "severity": "critical"}},
 
 		// https://issues.redhat.com/browse/OSD-3086
-		{Receiver: receiverPagerduty, MatchRE: map[string]string{"exported_namespace": alertmanager.PDRegex}},
+		// https://issues.redhat.com/browse/OSD-5872
+		{Receiver: receiverPagerduty, MatchRE: map[string]string{"exported_namespace": alertmanager.PDRegex}, Match: map[string]string{"prometheus": "openshift-monitoring/k8s"}},
```
jewzaam, line 156: This change only impacts alerts where `exported_namespace` matches our regex.  The `prometheus` label should be checked for EVERYTHING routed to the Receiver `receiverPagerduty`.

jewzaam (changes requested):

pkg/controller/secret/secret_controller.go
sam-nguyen7: added the label for all PD subroutes. PTAL @jewzaam

jewzaam (approved):

PR #142, opened 2021-02-23
Send PrometheusNotIngestingSamples for openshift-user-workload-monitoring to null receiver

fahlmant:
In 4.6, this alert fires by default as UWM has no scrape targets. Fixed in 4.7 but monitoring team recommended silencing the alerts.
https://issues.redhat.com/browse/OSD-6559

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -148,6 +148,9 @@ func createPagerdutyRoute() *alertmanager.Route {
 		// https://issues.redhat.com/browse/OSD-6327
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "CannotRetrieveUpdates"}},
 
+		//https://issues.redhat.com/browse/OSD-6559
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "PrometheusNotIngestingSamples", "namespace": "openshift-user-workload-monitoring"}},
+
 		// https://issues.redhat.com/browse/OSD-1922
 		{Receiver: receiverMakeItWarning, Match: map[string]string{"alertname": "KubeAPILatencyHigh", "severity": "critical"}},
 


iamkirkbater:
Love this.  +1 from me

jharrington22:
@fahlmant how do we remember to revert this when 4.7 comes around?

fahlmant:
@jharrington22 We can make a card and put it in the backlog. We'll have to wait until 4.6 is all gone so it will be a while.

jharrington22:
> @jharrington22 We can make a card and put it in the backlog. We'll have to wait until 4.6 is all gone so it will be a while.

Lets add a card and I'll +1 this.

pkg/controller/secret/secret_controller.go
```diff
@@ -148,6 +148,9 @@ func createPagerdutyRoute() *alertmanager.Route {
 		// https://issues.redhat.com/browse/OSD-6327
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "CannotRetrieveUpdates"}},
 
+		//https://issues.redhat.com/browse/OSD-6559
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "PrometheusNotIngestingSamples", "namespace": "openshift-user-workload-monitoring"}},
```
jewzaam, line 152: Why do we get any alerts for this namespace?  I expect we need to make sure UWM is up, but is there a specific set of alerts we should allow and discard the rest?

fahlmant: We're responsible for ensuring the resources in this namespace (prom pods, thanos pods) are running and working properly. Maybe we can make a list of Alerts we care about and scope it to that. For the short term, this will generate noise for every deployment by default, so we should silence this, then reevaluate other alerts from this NS IMO.

jewzaam: Ack.  Let's card up something to address this more long term but understood there's a tactical need.
/approve

fahlmant:
@jharrington22 https://issues.redhat.com/browse/OSD-6559

PR #143, opened 2021-03-08
OSD-6646: Manage osd-cluster-ready

2uasimojo:
With this commit, the operator:
- Manages the osd-cluster-ready Job (formerly managed by managed-cluster-config)
- Waits for osd-cluster-ready to complete before configuring PagerDuty.

[OSD-6646](https://issues.redhat.com/browse/OSD-6646)


2uasimojo:
I've got to leave this until next week.

**Current status:**
- Live testing is :+1: -- I've run through all the code paths at least once and they work as designed. However, I would like to do that again before we land this.
- Unit test is fixed up wrt the existing suite. Basically, I mocked out the new stuff in such a way that the behavior is as it was previously.
- **TODO:** Write unit tests for the new code, including adding flows through the secret reconciler for various permutations of `IsReady()`. If we're in a hurry, I suppose I can do that in a followon PR.
- I don't know what's going on with the coverage job. It has suddenly started failing intermittently (but more often than not) in multiple repositories. We suspect codecov.io has changed their API in some way. Someone needs to look into this eventually...

**Interdependencies:**
- https://github.com/openshift/osd-cluster-ready/pull/7 rips out the max cluster age and silence logic. I've already built and pushed this image to [quay](https://quay.io/repository/openshift-sre/osd-cluster-ready?tab=tags) as tags `v0.1.45-5bcea75` and `latest` (the latter being what this PR's Job config will pick up) -- so we don't technically need to merge that PR for this one to work.
- https://github.com/openshift/managed-cluster-config/pull/742 stops MCC from deploying osd-cluster-ready. I don't *think* there's any hard dependency order between that PR and this one:
  - c-am-o will treat a succeeded osd-cluster-ready Job the same whether it came from MCC or c-am-o: it will proceed to configure PagerDuty.
  - c-am-o will treat a running osd-cluster-ready Job the same whether it came from MCC or c-am-o: it will requeue waiting for it to complete before configuring PagerDuty.
  - If c-am-o sees a failed osd-cluster-ready Job, it will delete it and create its own. Since the MCC/hive-based Job is via a `CreateOnly` syncset, the only *minor* race would be if c-am-o deletes the Job and hive syncs it before c-am-o can recreate it. But there again, c-am-o should treat that Job properly.
  - If c-am-o gets in first and creates the Job, hive ought to no-op due to the `CreateOnly` setting.

2uasimojo:
> TODO: Write unit tests for the new code ~, including adding flows through the secret reconciler for various permutations of IsReady()~.

High-level Reconcile tests added. Still need UT for the `readiness` module.

2uasimojo:
Closing in favor of #147. TL;DR: c-am-o isn't the right owner for the osd-cluster-ready Job. #147 keeps the controller logic, but still relies on MCC to deploy the actual Job.

PR #146, opened 2021-03-10
Disable Alerts for Prometheus Remote Writes

arjunrn:
Remote writes of metrics data to Observatorium can occasionally fail due to an outage. This triggers an alert in-cluster. To prevent alert fatigue let's disable the alerts related to remote writes.

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -151,6 +151,11 @@ func createPagerdutyRoute() *alertmanager.Route {
 		//https://issues.redhat.com/browse/OSD-6559
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "PrometheusNotIngestingSamples", "namespace": "openshift-user-workload-monitoring"}},
 
+		//https://issues.redhat.com/browse/OSD-6704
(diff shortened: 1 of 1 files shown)


jharrington22 (approved):
/lgtm

jharrington22:
@cblecker @lisa @jewzaam this PR is to silence prometheus alerts for remote writes to OSD's Observatorium. These alerts haven't been seen before because remote write is not enabled on clusters. As the Observatorium team is scaling up to handle the load we want to initially silence these alerts to ensure we don't fatigue SRE. We have a plan to reenable these alerts https://issues.redhat.com/browse/OSD-6709

jewzaam:
Before removing hold please add a card tracking undoing this with condition being Observatorium has an SLA.  We care about these metrics for reporting status of clusters via OCM at a minimum.  Once the Observatorium service is better supported we need to care about when it's offline, at least so we can ship those alerts to the Observatorium team.

/hold
/approve

jewzaam:
Guess @jharrington22 provided what I asked for right when I asked.  Thanks!  Removing hold.  :shipit: 

/hold cancel

PR #147, opened 2021-03-17
OSD-6646: Wait for ready job before configuring PD

2uasimojo:
This commit coordinates with the osd-cluster-ready Job to avoid throwing alerts while the cluster is provisioning. We configure DMS, but wait to configure PagerDuty until that Job has completed (success or failure). We wait a maximum of two hours (configurable via the `MAX_CLUSTER_AGE_MINUTES` env var) for the Job to complete, at which point we configure PagerDuty anyway.

[OSD-6646](https://issues.redhat.com/browse/OSD-6646)


jharrington22:
/lgtm

Major difference here with what we are currently doing in osd-cluster-ready is that we're now querying prom using Go libs which is great and looks ok to me.

@2uasimojo testing in stage shows this is ok? Have you been able to test the `MAX_CLUSTER_AGE_MINUTES` threshold?

2uasimojo:
> @2uasimojo testing in stage shows this is ok? Have you been able to test the `MAX_CLUSTER_AGE_MINUTES` threshold?

Yes and yes. All the code paths are :+1:

PR #150, opened 2021-03-26
Update origin-operator-registry image tag

cblecker:
fixes [OSD-6829](https://issues.redhat.com/browse/OSD-6829)

### DIFF
--- a/build/Dockerfile.catalog_registry
+++ b/build/Dockerfile.catalog_registry
@@ -1,4 +1,4 @@
-FROM quay.io/openshift/origin-operator-registry:4.5
+FROM quay.io/openshift/origin-operator-registry:latest
 
 ARG SRC_BUNDLES
 


build/Dockerfile.catalog_registry
```diff
@@ -1,8 +1,8 @@
-FROM quay.io/openshift/origin-operator-registry:4.5
+FROM quay.io/openshift/origin-operator-registry:latest
```
NautiluX, line 1: I'd prefer using a specific tag instead of latest here for the following reasons:
* Using latest to avoid CVEs gives a false sense of security, as still a PR to the repo needs to be merged to have the CD create and deploy a new operator image.
* Unrelated PRs will use newer operator-registry images which may break the operator and debugging becomes a nightmare because it's not obvious the image changed.
Rather than setting this to latest on all operators, I'd prefer a pipeline that opens PRs when new images are released against all repositories. Until that's in place, l suggest we just use the latest currently available image (4.9 according to [this PR](https://github.com/openshift/gcp-project-operator/pull/134)).

NautiluX (changes requested):

build/Dockerfile.catalog_registry
lisa: I agree with this. 👍

lisa: Furthermore, it takes some effort to determine what base image has been used for a certain deployed version. By pinning to a specific static tag we have greater predictability.

PR #165, opened 2021-06-01
Stop filtering based on `namespace` label

RiRa12621:
only core namespaces are watched anyway by cluster monitoring.
Filtering solely based on alerts originating from the main prometheus is good enough and ensures we get all alerts.

With the current filtering we are missing alert like https://github.com/openshift/machine-config-operator/blob/release-4.6/install/0000_90_machine-config-operator_01_prometheus-rules.yaml that don't have a namespace label but we definitely care about

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -170,9 +170,9 @@ func createPagerdutyRoute() *alertmanager.Route {
 		{Receiver: receiverMakeItWarning, Match: map[string]string{"alertname": "KubeAPILatencyHigh", "severity": "critical"}},
 
 
-		{Receiver: receiverPagerduty, MatchRE: map[string]string{Match: map[string]string{"prometheus": "openshift-monitoring/k8s"}},
+		{Receiver: receiverPagerduty, Match: map[string]string{"prometheus": "openshift-monitoring/k8s"}},
 		// general: route anything in core namespaces to PD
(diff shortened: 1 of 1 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -169,18 +169,10 @@ func createPagerdutyRoute() *alertmanager.Route {
 		// https://issues.redhat.com/browse/OSD-1922
 		{Receiver: receiverMakeItWarning, Match: map[string]string{"alertname": "KubeAPILatencyHigh", "severity": "critical"}},
 
-		// https://issues.redhat.com/browse/OSD-3086
-		// https://issues.redhat.com/browse/OSD-5872
-		{Receiver: receiverPagerduty, MatchRE: map[string]string{"exported_namespace": alertmanager.PDRegex}, Match: map[string]string{"prometheus": "openshift-monitoring/k8s"}},
+
+		{Receiver: receiverPagerduty, Match: map[string]string{"prometheus": "openshift-monitoring/k8s"}},
```
cblecker, line 173: This line is the sane as L175

pkg/controller/secret/secret_controller.go
```diff
@@ -169,18 +169,10 @@ func createPagerdutyRoute() *alertmanager.Route {
...
 		// general: route anything in core namespaces to PD
```
cblecker, line 174: This comment isn't accurate anymore

cblecker (commented):
Have we tested this to ensure we don't get customer workload alerts?

RiRa12621:
/close
Superseded by #166

PR #166, opened 2021-06-04
OSD-7438 remove namespace filter

yithian:


yithian:
@cblecker you mentioned that you had some concerns about this, can you add them here?

PR #168, opened 2021-06-29
dev null pipelines

dofinn:
https://issues.redhat.com/browse/OSD-7631 

Alerts coming from this namespace are from tekton/operator-hub. 

SRE should not be alerted from this namespace.


NautiluX (commented):
/lgtm

PR #170, opened 2021-07-08
Added alerts to dev/null receiver

rafael-azevedo:
This pr adds the following alerts with severity warning to dev/null receiver per https://issues.redhat.com/browse/OSD-7671

-- FluentdQueueLengthBurst
-- FailingOperator

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -198,6 +198,10 @@ func createPagerdutyRoute() *alertmanager.Route {
 		// https://issues.redhat.com/browse/OSD-6821
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "PrometheusBadConfig", "namespace": "openshift-user-workload-monitoring"}},
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "PrometheusDuplicateTimestamps", "namespace": "openshift-user-workload-monitoring"}},
+
+		//https://issues.redhat.com/browse/OSD-7671
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "FluentdQueueLengthBurst", "namespace": "openshift-logging", "severity": "warning"}},
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "FailingOperator", "namespace": "openshift-operator-lifecycle-manager", "severity": "warning"}},
 	}
 
 	return &alertmanager.Route{


pkg/controller/secret/secret_controller.go
```diff
@@ -198,6 +198,10 @@ func createPagerdutyRoute() *alertmanager.Route {
 		// https://issues.redhat.com/browse/OSD-6821
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "PrometheusBadConfig", "namespace": "openshift-user-workload-monitoring"}},
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "PrometheusDuplicateTimestamps", "namespace": "openshift-user-workload-monitoring"}},
+
+		//https://issues.redhat.com/browse/OSD-7671
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "FluentdQueueLengthBurst", "namespace": "openshift-logging", "severity": "warning"}},
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "FailingOperator", "namespace": "openshift-operator-lifecycle-manager", "severity": "warning"}},
```
fahlmant, line 204: Do you an example of this alert?

rafael-azevedo: https://redhat.pagerduty.com/incidents/PBCRZ2U

https://redhat.pagerduty.com/incidents/PJF9J4M

jharrington22: Is https://redhat.pagerduty.com/incidents/PJF9J4M not a valid alert? route-monitor-operator in this case didn't install?

rafael-azevedo: > 
> 
> Is https://redhat.pagerduty.com/incidents/PJF9J4M not a valid alert? route-monitor-operator in this case didn't install?

We have alerts in place that grab this as a critical I believe or covers this use case

jharrington22: No problem routing the first one to /dev/null has there isn't much we can do with that alert probably a candidate to take up with the logging team - maybe a card? cc/ @RiRa12621 

The second one though I think we want that, though it is a warning you are right. It hasn't cleared so something is up with OLM.

RiRa12621:
@rafael-azevedo can you link the corresponding upstream BZs please?

pkg/controller/secret/secret_controller.go
RiRa12621: Happy to get the noise off us if it is not actionable but there should be an upstream BZ or actually two since there are two different alerts.

https://github.com/openshift/configure-alertmanager-operator/pull/170#issuecomment-876440066

jharrington22: @RiRa12621 I believe the second one is actionable as its pointing to an issue with OLM and I see its still not resolved so we should look into it. 

Can we split this PR into two so we can merge the Fluentd routing once a BZ is created?

RiRa12621: From the slack conversation, it looks like that's happening already.

mmazur: > Is https://redhat.pagerduty.com/incidents/PJF9J4M not a valid alert? route-monitor-operator in this case didn't install?

This also generated a KubeContainer Waiting WARNING, so if Primary had time for checking Low alerts, they would've investigated the cluster anyway.

jharrington22:
This is closed by #171 and #172

PR #173, opened 2021-07-19
[OSD-7437] Disable KubeJobFailed alert for image-pruner

dastergon:
After dealing with multiple failed job alerts for `image-pruner`, we realized that currently, a single failed image-pruner cronjob isn't indicative of an ongoing issue. This PR disables the upstream `KubeJoFailed` alert for image-pruner. I will send a PR to managed-cluster-config for a custom alert for image-pruner so that an alert won't get triggered until 2 jobs have failed.

PTAL 

See: https://issues.redhat.com/browse/OSD-7437

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -149,6 +149,8 @@ func createPagerdutyRoute() *alertmanager.Route {
 		{Receiver: receiverNull, MatchRE: map[string]string{"namespace": alertmanager.PDRegexLP}, Match: map[string]string{"alertname": "TargetDown"}},
 		// https://issues.redhat.com/browse/OSD-5544
 		{Receiver: receiverNull, MatchRE: map[string]string{"job_name": "^elasticsearch.*"}, Match: map[string]string{"alertname": "KubeJobFailed", "namespace": "openshift-logging"}},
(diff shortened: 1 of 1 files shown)


RiRa12621:
/hold
https://github.com/openshift/managed-cluster-config/pull/882
needs to definitely be merged beforehand. otherwise we'll have an alert on cluster that's blocking upgrades and we don't know, because we don't forward

bdematte:
openshift/managed-cluster-config#882 has been merged. 
However, would be good to have https://github.com/openshift/managed-cluster-config/pull/883 merged before removing this one.

RiRa12621:
/unhold

We have https://github.com/openshift/managed-cluster-config/blob/master/deploy/sre-prometheus/100-sre-pruning.PrometheusRule.yaml that should cover most of our problems already.
Still needs lgtm and approved but my concerns are adressed.

cblecker:
/hold

https://github.com/openshift/managed-cluster-config/blob/master/deploy/sre-prometheus/100-sre-pruning.PrometheusRule.yaml is for the sre pruning jobs in the `openshift-sre-pruning` namespace. It doesn't cover the image registry pruner in the `openshift-image-registry` namespace as highlighted in https://github.com/openshift/managed-cluster-config/pull/883

PR #175, opened 2021-07-23
Move alert ignore earlier in the func to actually work

mmazur:
This corrects #171 and should actually make the ignore work.

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -172,6 +172,10 @@ func createPagerdutyRoute() *alertmanager.Route {
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "PrometheusRemoteWriteBehind", "namespace": "openshift-monitoring"}},
 
 		{Receiver: receiverNull, Match: map[string]string{"namespace": "openshift-gitops"}},
+
+		//https://issues.redhat.com/browse/OSD-7671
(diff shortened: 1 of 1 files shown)


georgettica:
I can 
/lgtm
but isn't there an order we need to enforce now to make this issue not happen even again
/hold
so we can discuss this

fahlmant:
What's the reason it needed to be moved up to work?

mmazur:
Seems to me the alert [example](https://redhat.pagerduty.com/incidents/PMPZ342) is getting caught by one of the rules below, which is why it's not working:

```
{Receiver: receiverPagerduty, MatchRE: map[string]string{"exported_namespace": alertmanager.PDRegex}, Match: map[string]string{"prometheus": "openshift-monitoring/k8s"}},
{Receiver: receiverPagerduty, MatchRE: map[string]string{"namespace": alertmanager.PDRegex}, Match: map[string]string{"exported_namespace": "", "prometheus": "openshift-monitoring/k8s"}},
{Receiver: receiverPagerduty, Match: map[string]string{"job": "fluentd", "prometheus": "openshift-monitoring/k8s"}},
```

hence this patch to correct it.

@georgettica I don't know what you're referring to.

georgettica:
If this rule is caught by something else, is that a one off? Or should we invest time in making sure these things (metrics being caught by other regexes) will always work?

RiRa12621:
I'd say that's a longer effort and should be carded up a follow up work

mmazur:
Soooo....?

RiRa12621:
As mentioned on slack before, please create the follow up card before we merge this.

mmazur:
Done

PR #182, opened 2021-09-03
Explicitly redirect auth operator ClusterOperatorDown to null

fahlmant:
https://issues.redhat.com/browse/OSD-8320

There is an alert suppression configuration that prevents ClusterOperatorDown from firing if ClusterOperatorDegraded is also firing. Recently we've seen some cases where only ClusterOperatorDown is firing for the auth operator, meaning it does not get suppressed. This explicitly redirects those alerts to null


RiRa12621:
/hold
/lgtm
follow up question in the ticket, unhold to merge once that's cleared

yithian:
The rules themselves seem sane. There should be test cases for them added to https://github.com/openshift/configure-alertmanager-operator/blob/master/pkg/controller/secret/secret_controller_test.go though

PR #183, opened 2021-09-06
[OSD-8337] Ignore openshift-storage namespace from alerting

chopraapooja:


RiRa12621:
/retitel Jira OSD-8337: Ignore openshift-storage namespace from alerting

PR #186, opened 2021-09-08
OSD-7842 - Only route alerts to PagerDuty that originate from a managed namespace

tnierman:
Completes https://issues.redhat.com/browse/OSD-7842

These changes utilize the newly generated [managed-namespaces configMap](https://github.com/openshift/managed-cluster-config/blob/master/deploy/osd-managed-resources/managed-namespaces.ConfigMap.yaml) to only route alerts from OSD namespaces to PagerDuty.

### Alerts
- Alert fires from namespace present in `managed-namespaces` CM -> Alert appears in PD
- Alert fires from namespace not present in `managed-namespaces` CM -> Alert does not appear in PD 
- Alert is firing in a managed namespace, and namespace is removed from `managed-namespaces` CM -> Alert remains in PagerDuty
- Alert is firing in unmanaged namespace, and namespace is added to `managed-namespaces` CM -> Alert appears in PD

### Namespaces
- `managed-namespaces` ConfigMap is invalid, empty, or removed -> Alerts are routed to PagerDuty from all `^openshift-.*` namespaces

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -34,6 +34,10 @@ const (
 
 	cmKeyManagedNamespaces = "managed_namespaces.yaml"
 
+	cmKeyOCPNamespaces = "managed_namespaces.yaml"
+
+	cmKeyAddonsNamespaces = "managed_namespaces.yaml"
+
 	secretNamePD = "pd-secret"
 
 	secretNameDMS = "dms-secret"
@@ -41,6 +45,10 @@ const (
 	secretNameAlertmanager = "alertmanager-main"
 
 	cmNameManagedNamespaces = "managed-namespaces"
+
+	cmNameOCPNamespaces = "ocp-namespaces"
+
+	cmNameAddonsNamespaces = "addons-namespaces"
 
 	// anything routed to "null" receiver does not get routed to PD
 	receiverNull = "null"
@@ -60,6 +68,12 @@ const (
 	// global config for PagerdutyURL
 	pagerdutyURL = "https://events.pagerduty.com/v2/enqueue"
 )
+
+var defaultNamespaces = []string{
+	alertmanager.PDRegexOS,
+	alertmanager.PDRegexLP,
+	alertmanager.PDRegexKube,
+}
 
 var _ reconcile.Reconciler = &ReconcileSecret{}
 
@@ -111,7 +125,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 }
 
 // createPagerdutyRoute creates an AlertManager Route for PagerDuty in memory.
-func createPagerdutyRoute(managedNamespaceList []string) *alertmanager.Route {
+func createPagerdutyRoute(namespaceList []string) *alertmanager.Route {
 	// order matters.
 	// these are sub-routes.  if any matches it will not continue processing.
 	// 1. route anything we want to silence to "null"
@@ -229,7 +243,7 @@ func createPagerdutyRoute(managedNamespaceList []string) *alertmanager.Route {
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "CsvAbnormalFailedOver2Min", "exported_namespace": "redhat-ods-operator"}},
 	}
 
-	for _, namespace := range managedNamespaceList {
+	for _, namespace := range namespaceList {
 		pagerdutySubroutes = append(pagerdutySubroutes, []*alertmanager.Route{
 			// https://issues.redhat.com/browse/OSD-3086
 			// https://issues.redhat.com/browse/OSD-5872
@@ -325,7 +339,7 @@ func createWatchdogReceivers(watchdogURL string) []*alertmanager.Receiver {
 }
 
 // createAlertManagerConfig creates an AlertManager Config in memory based on the provided input parameters.
-func createAlertManagerConfig(pagerdutyRoutingKey, watchdogURL, clusterID string, managedNamespaceList []string) *alertmanager.Config {
+func createAlertManagerConfig(pagerdutyRoutingKey, watchdogURL, clusterID string, namespaceList []string) *alertmanager.Config {
 	routes := []*alertmanager.Route{}
 	receivers := []*alertmanager.Receiver{}
 
@@ -335,7 +349,7 @@ func createAlertManagerConfig(pagerdutyRoutingKey, watchdogURL, clusterID string
 	}
 
 	if pagerdutyRoutingKey != "" {
-		routes = append(routes, createPagerdutyRoute(managedNamespaceList))
+		routes = append(routes, createPagerdutyRoute(namespaceList))
 		receivers = append(receivers, createPagerdutyReceivers(pagerdutyRoutingKey, clusterID)...)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -529,11 +555,35 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
...
 		watchdogURL = readSecretKey(r, &request, secretNameDMS, secretKeyDMS)
 	}
 
+	if namespaceCMExists {
+		log.Info("INFO: Managed namespace configMap exists")
+		var managedNamespacesConfig alertmanager.ManagedNamespacesConfig
+		rawManagedNamespaces := readCMKey(r, &request, cmNameManagedNamespaces, cmKeyManagedNamespaces)
+
+		err := yaml.Unmarshal([]byte(rawManagedNamespaces), &managedNamespacesConfig)
+		if (err != nil) || (len(managedNamespacesConfig.Resources.Namespace) == 0) {
```
tnierman, line 570: `yaml.Unmarshal` doesn't throw an error if the `managedNamespacesConfig` struct doesn't match, checking if the namespace list is empty seems to be the easiest way to determine if something is off with the yaml

tnierman: Since this should be noted for the future, I added a comment clarifying this.

pkg/controller/secret/secret_controller.go
```diff
@@ -579,6 +630,39 @@ func secretInList(name string, list *corev1.SecretList) bool {
 	return false
 }
 
+// cmInList takes the name of ConfigMap, and a list of ConfigMaps, and returns a Bool
+// indicating if the name is present in the list
+func cmInList(name string, list *corev1.ConfigMapList) bool {
```
dustman9000, line 635: Can you add some unit tests around these functions in secret_controller_test.go?

tnierman: [Unit tests added](https://github.com/openshift/configure-alertmanager-operator/pull/186/files#diff-3f3b1c34e4d392411c2522d11db21d7d89b9603de69a11f63686c664558bb375R387-R429)

dustman9000: Thanks!

pkg/controller/secret/secret_controller.go
```diff
@@ -529,11 +555,36 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
...
 	}
 
+	if namespaceCMExists {
+		log.Info("INFO: Managed namespace configMap exists")
+		var managedNamespacesConfig alertmanager.ManagedNamespacesConfig
+		rawManagedNamespaces := readCMKey(r, &request, cmNameManagedNamespaces, cmKeyManagedNamespaces)
+
+		err := yaml.Unmarshal([]byte(rawManagedNamespaces), &managedNamespacesConfig)
+		// yaml.Unmarshal does not throw an error if managedNamespacesConfig does not match -> instead, check that namespace list is empty
+		if (err != nil) || (len(managedNamespacesConfig.Resources.Namespace) == 0) {
```
dustman9000, line 571: Just a minor observation that we could DRY this code up a bit and have `managedNamespaceList` be set to the default (`redhat-*`, `kube*`, and `openshift-*`) and only check for err == nil here. Appending the managed namespaces would be the exception and you wouldn't need all of these else blocks.

dustman9000:
Nice work @tnierman! Just some minor comments/questions, but nothing major to hold this from merging. Thanks!
/lgtm

pkg/controller/secret/secret_controller.go
```diff
@@ -517,6 +541,8 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 	// do the work! collect secret info for PD and DMS
 	pagerdutyRoutingKey := ""
 	watchdogURL := ""
+	managedNamespaceList := []string{alertmanager.PDRegexLP, alertmanager.PDRegexKube}
```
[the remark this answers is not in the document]
tnierman: Looks like there's still one object we might want to alert on in `^kube.*` namespaces:
```
$ oc get namespaces | grep -e ^kube.*
kube-node-lease                                    Active   29h
kube-public                                        Active   29h
kube-system                                        Active   29h
```

```
$ oc get all -n kube-system
NAME              TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)                        AGE
service/kubelet   ClusterIP   None         <none>        10250/TCP,10255/TCP,4194/TCP   29h
```

```
$ oc get all -n kube-public
No resources found in kube-public namespace.
```

```
$ oc get all -n kube-node-lease
No resources found in kube-node-lease namespace.
```

By default, I don't think OSD clusters have any `^redhat-.*` namespaces:
```
$ oc get namespaces | grep -e ^redhat-.*
$
```

Given this, I'd be in favor of not alerting on `^redhat-.*` namespaces, but I think it's desirable for us to continue to monitor `^kube.*` namespaces.

pkg/controller/secret/secret_controller.go
```diff
@@ -465,12 +479,13 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 	reqLogger := log.WithValues("Request.Name", request.Name)
 	reqLogger.Info("Reconciling Secret")
 
-	// This operator is only interested in the 3 secrets listed below. Skip reconciling for all other secrets.
+	// This operator is only interested in the 3 secrets & 1 configMap listed below. Skip reconciling for all other objects.
```
fahlmant, line 482: Minor nit, but with the watch of confimaps, 480 should also be updated

pkg/controller/secret/secret_controller.go
```diff
@@ -465,12 +479,13 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
...
 	// TODO: Filter these with a predicate instead
 	switch request.Name {
 	case secretNamePD:
 	case secretNameDMS:
 	case secretNameAlertmanager:
+	case cmNameManagedNamespaces:
```
fahlmant, line 488: Now that there are multiple types being handled here, a type check, at least for the once case, should be added.

pkg/controller/secret/secret_controller.go
```diff
@@ -497,6 +512,15 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
...
+	if err != nil {
+		log.Error(err, "Unable to list configMaps")
+	}
+
+	// Check for the presence of specific configMaps.
+	namespaceCMExists := cmInList(cmNameManagedNamespaces, cmList)
+
 	// Get the secret from the request.  If it's a secret we monitor, flag for reconcile.
 	instance := &corev1.Secret{}
 	err = r.client.Get(context.TODO(), request.NamespacedName, instance)
```
fahlmant, line 526: Does this work if the type of object that triggered the reconcile is a configmap?

pkg/controller/secret/secret_controller.go
tnierman: Looks like `request` is a `reconcile.Request` object, which only contains a name & namespace, so I don't know if a type check is feasible. 

The good news is that even if there is both a secret and configmap named after `cmNameManagedNamespaces`, the reconcile loop should still ignore the secret and only interact with the configmap, unless I'm misunderstanding the intention here.

pkg/controller/secret/secret_controller.go
tnierman: Yeah, this function will return true if, any time the reconcile loop is run, there exists a configmap named after the variable `cmNameManagedNamespaces`. 

The `reconcile` loop now runs every time a configmap or secret changes, before it only would run if a secret changes.

pkg/controller/secret/secret_controller.go
tnierman: Changed `"Reconciling Secret"` -> `"Reconciling Object"`

pkg/controller/secret/secret_controller.go
tnierman: Ignore might be the wrong word, but I don't think it would have any ill-effect. 

In a case where a configmap name matches a CAMO secret name and triggers a reconcile (or vice-versa), the request name is only used to lookup the correct objects and act on them. In this case, the reconcile _is_ called unnecessary, but it shouldn't result in a different `alertmanager-main` secret, since the same objects are used to generate it each time.

pkg/controller/secret/secret_controller.go
tnierman: Ah, I just realized you meant line 526, not 522, my bad. 
The block just continues if it can't find a secret with the request name, so it doesn't affect reconciles with the `managed-namespaces` ConfigMap. I've added a corresponding block to check for ConfigMap errors as well.

pkg/controller/secret/secret_controller.go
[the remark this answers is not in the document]
pbergene: ^redhat-.* contain managed applications.  I'm not sure if there is any reliance on routing alerts via the cluster prometheus, but worth checking with @hlipsig @apahim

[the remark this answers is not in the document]
hlipsig: +1 redhat-* for add-ons.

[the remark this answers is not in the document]
dustman9000: @cblecker - Can you confirm that we need to alert on add-ons? My understanding was that we do not want to alert on layered products/add-ons under the platform alertmanager. The thinking behind this change was to have an explicit list, not a regex.

cblecker:
The configmap (https://github.com/openshift/managed-cluster-config/blob/master/deploy/osd-managed-resources/managed-namespaces.ConfigMap.yaml) needs to be more exhaustive before we can merge this. We need to have all the OCP namespaces in there.

We also need to ensure the following alerts will still fire/route after this:
https://github.com/openshift/ops-sop/blob/master/v4/alerts/RHMIUpgradeExpectedDurationExceeded.md
https://github.com/openshift/ops-sop/blob/master/v4/alerts/RHOAMOperatorInstallDelayed.md
https://github.com/openshift/ops-sop/blob/master/v4/alerts/RHOAMUpgradeExpectedDurationExceeded.md

pkg/controller/secret/secret_controller.go
```diff
@@ -448,6 +462,71 @@ func createAlertManagerConfig(pagerdutyRoutingKey, watchdogURL, clusterID string
...
+	}
+
+	// If we're able to access the namespace list in managed-namespaces, then use the list's namespaces
+	for _, namespace := range managedNamespacesConfig.Resources.Namespace {
+		managedNamespaceList = append(managedNamespaceList, "^" + namespace.Name + "$")
+	}
+	return managedNamespaceList
+}
+
+//func (r *ReconcileSecret) parseSecrets(request reconcile.Request, clusterReady bool) (pagerdutyRoutingKey string, watchdogURL string, err error) {
```
fahlmant, line 495: Can this be removed?

fahlmant (approved):
/lgtm

PR #187, opened 2021-09-13
Added watchdog alerts first

rafael-azevedo:
A namespace was added to the watch dog alert that is matching rules prior to the watchdog rule on the config. I changed the order in which the config is made so that watchdog is matched first. 

https://bugzilla.redhat.com/show_bug.cgi?id=2003181

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -306,14 +306,14 @@ func createAlertManagerConfig(pagerdutyRoutingKey, watchdogURL, consoleUrl strin
 	routes := []*alertmanager.Route{}
 	receivers := []*alertmanager.Receiver{}
 
-	if pagerdutyRoutingKey != "" {
-		routes = append(routes, createPagerdutyRoute())
-		receivers = append(receivers, createPagerdutyReceivers(pagerdutyRoutingKey, consoleUrl, clusterID)...)
-	}
-
 	if watchdogURL != "" {
 		routes = append(routes, createWatchdogRoute())
 		receivers = append(receivers, createWatchdogReceivers(watchdogURL)...)
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -278,7 +278,7 @@ func createWatchdogRoute() *alertmanager.Route {
 	return &alertmanager.Route{
 		Receiver:       receiverWatchdog,
 		RepeatInterval: "5m",
-		Match:          map[string]string{"alertname": "Watchdog"},
+		Match:          map[string]string{"alertname": "Watchdog", "namespace": "openshift-monitoring"},
```
cblecker, line 281: I don't think this is needed. This will break versions that don't have this label. Matching on the alertname is good enough.

pkg/controller/secret/secret_controller.go
```diff
@@ -306,16 +306,16 @@ func createAlertManagerConfig(pagerdutyRoutingKey, watchdogURL, consoleUrl strin
...
 	if watchdogURL != "" {
 		routes = append(routes, createWatchdogRoute())
 		receivers = append(receivers, createWatchdogReceivers(watchdogURL)...)
 	}
 
+	if pagerdutyRoutingKey != "" {
+		routes = append(routes, createPagerdutyRoute())
+		receivers = append(receivers, createPagerdutyReceivers(pagerdutyRoutingKey, consoleUrl, clusterID)...)
+	}
+
```
cblecker, line 318: Ugh. Yeah, ordering. This change lgtm

pkg/controller/secret/secret_controller.go
fahlmant: ah good point. I suggested this so blame me :)

PR #191, opened 2021-10-11
OSD-8576 remove fluentd-q sev to catch all

dofinn:


dofinn:
Was working on miss-information here. Closing.

PR #192, opened 2021-10-12
OSD-8665 devnull vol capacity warnings

dofinn:
This is currently a NO OP for SRE. It should not reach pager.


NautiluX (commented):
/lgtm

PR #193, opened 2021-10-14
Ignores Alerts for CsvAbnormalFailedOver2Min for redhat-ods-operator 

iamkirkbater:
Since the issue is bubbled up to the customer already, there's no reason we should get this alert. [OSD-8689](https://issues.redhat.com/browse/OSD-8689)


fahlmant (approved):
/approve

PR #200, opened 2021-11-21
Route alerts with improper severity to 'null'

cblecker:
If alerts have malformed severities, they will cause 400 errors from the PD api, and errors from alertmanager.

fixes [OSD-9062](https://issues.redhat.com/browse/OSD-9062)

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -241,6 +241,9 @@ func createPagerdutyRoute(namespaceList []string) *alertmanager.Route {
 
 		// https://issues.redhat.com/browse/OSD-8689
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "CsvAbnormalFailedOver2Min", "exported_namespace": "redhat-ods-operator"}},
+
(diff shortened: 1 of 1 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -241,6 +241,9 @@ func createPagerdutyRoute(namespaceList []string) *alertmanager.Route {
 
 		// https://issues.redhat.com/browse/OSD-8689
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "CsvAbnormalFailedOver2Min", "exported_namespace": "redhat-ods-operator"}},
+
+		// https://issues.redhat.com/browse/OSD-9062
+		{Receiver: receiverNull, Match: map[string]string{"severity": "alert"}},
```
dofinn, line 246: Is there anyway we can inverse this to != "warning/critical/info" ?

cblecker: Golang doesn't support negative lookarounds in its regex matching, so no, not easily.

PR #202, opened 2021-12-07
Also catch severity=error

iamkirkbater:
This alert also fires as an error level alert, so let's route this to /dev/null for any severity level.


fahlmant (approved):
/approve

PR #203, opened 2021-12-10
OSD-9239 Add a deployment parameter OLM template for fedramp

luis-falcon:
This closes [OSD-9239](https://issues.redhat.com/browse/OSD-9239):

added to the OLM template, such that AppSRE can deploy the operator in different govcloud vs. non-govcloud environments.

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -3,6 +3,8 @@ package secret
 import (
 	"context"
 	"fmt"
+	"os"
+	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -595,6 +598,12 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 	reqLogger := log.WithValues("Request.Name", request.Name)
 	reqLogger.Info("Reconciling Object")
 
+	if len(os.Getenv("FEDRAMP")) == 0 {
```
cblecker, line 601: As opposed to comparing length, there is a LookupEnv function: https://pkg.go.dev/os#LookupEnv

it also allows us to keep the logic in one place, as opposed to us having it here and on L31

PR #204, opened 2021-12-14
OSD-9027 - Include request name in logs

tnierman:
Completes: https://issues.redhat.com/browse/OSD-9027

---

Adds the request name to the `secret_controller` logs in order to provide more context for debugging

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -19,6 +19,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/go-logr/logr"
 	"github.com/openshift/configure-alertmanager-operator/config"
 	"github.com/openshift/configure-alertmanager-operator/pkg/metrics"
 	"github.com/openshift/configure-alertmanager-operator/pkg/readiness"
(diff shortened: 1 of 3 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -16,6 +16,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/go-logr/logr"
```
cblecker, line 19: This doesn't look right

pkg/controller/secret/secret_controller.go
```diff
@@ -26,6 +27,8 @@ import (
 )
 
 var log = logf.Log.WithName("secret_controller")
+var reqLogger logr.Logger
```
cblecker, line 30: Yeah this isn't right.. you're just instantiating a new log without actually specifying what you want to attach in.

Inside the reconcile loop, something like this would work:
```
reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
```

cblecker (changes requested):
/hold

pkg/controller/secret/secret_controller.go
tnierman: What specifically doesn't look right? The same package is used in other operators we maintain, ie- [here](https://github.com/openshift/managed-velero-operator/blob/master/pkg/controller/velero/velero.go#L22).

Certainly open to removing if there's a way to implement these changes without the additional dependency though!

fahlmant (commented):
Generally looks good, waiting on comment from Christoph

PR #205, opened 2021-12-20
OSD-9383 route sev=none to dev/null

dofinn:
Watchdog attempts to send an alert to PD with severity = none. This causes AM to raise an alert as PD rejects this. 

```
Warning Alerts:
{
  "alertname": "AlertmanagerReceiversNotConfigured",
  "job": null,
  "namespace": "openshift-monitoring",
  "exported_namespace": null,
  "pod": null,
  "service": null,
  "alertstate": "firing"
}
```

This PR ensures alerts with severity=none are routed to /dev/null to avoid "noise" on the pager.


cblecker:
/hold

This doesn't make sense to me. PD shouldn't ever get this alert as it should be routed to DMS.

dofinn:
Strange, I encounted this in staging on two 4.9 clusters yesterday. Today, I am unable to find any - appears to be fallout from FedRamp efforts. Closing

PR #207, opened 2021-12-22
put severity=alert before cluster=elasticsearch

feichashao:
After https://github.com/openshift/configure-alertmanager-operator/pull/200 being promoted, we still see AlertmanagerFailedToSendAlerts .

The order in the AM routes matters. The `severity=alert` is caught by `cluster=elasticsearch` which sends to PD.
https://github.com/openshift/configure-alertmanager-operator/blob/b2122556e52efba467f454e0ee25815b702f0bc5/pkg/controller/secret/secret_controller.go#L235

Moving the `severity=alert` route before `cluster=elasticsearch` to make it work as expected.


feichashao:
/hold
due to code freeze.

feichashao:
cc @cblecker @mrbarge

PR #209, opened 2022-01-14
OSD-9243 minor change: adding instruction to route to Error severity

ritmun:
added instruction to reduce alert urgency per the spike performed in [OSD-9243](https://issues.redhat.com/browse/OSD-9243)


fahlmant (approved):
/lgtm

PR #211, opened 2022-02-04
Improve fedramp check

sam-nguyen7:
Move fedramp check from reconcile to only check once at operator start up.


fahlmant (commented):
Please add a description :)

PR #212, opened 2022-02-08
OSD-9060 - Remove extraneous alerts

tnierman:
Completes https://issues.redhat.com/browse/OSD-9060

---

Removes PagerDuty routes for namespaces that are not present in the [managed-resources configmaps](https://github.com/openshift/managed-cluster-config/tree/master/deploy/osd-managed-resources) and route to the null receiver. 

These routes can safely be removed, since alerts from namespaces not explicitly defined in the managed-resources configmaps should not be routed to SRE.


lisa:
@tnierman I noticed the link points to your fork of [openshift/managed-cluster-config](https://github.com/openshift/managed-cluster-config). Was that intended?

tnierman:
@lisa Nope, thanks for pointing that out! I updated the link to point to openshift/managed-cluster-config

fahlmant (approved):
/approve

fahlmant (commented):
/lgtm

PR #216, opened 2022-02-28
OSD-10045 Add support for OCM Agent forwarding

mrbarge:
This PR adds support to CAMO to route alerts to the `ocm-agent-operator` [0]

More detail concerning the design of this change can be found linked in the associated ticket [OSD-10045](https://issues.redhat.com/browse/OSD-10045), but a summary is included below.

## Changes

CAMO will now watch for a new `ConfigMap` in `openshift-monitoring` named `ocm-agent` 

If this ConfigMap is present, CAMO will configure AlertManager to route alerts to the OCM Agent service running on the cluster (a new Receiver, in AlertManager terminology). The OCM Agent service URL is specified in the ConfigMap.

OCM Agent Operator (OAO) manages the creation/removal of the ConfigMap. If an OCM Agent CR is created on-cluster, OAO creates the ConfigMap. If the CR is removed, the ConfigMap is similarly removed by OAO. If the ConfigMap is changed, OAO will revert it to its expected state.

This PR also establishes two routing rules concerning OCM Agent:

- If an alert contains the label `send_managed_notification`, it is routed to the OCM Agent receiver.
- If an alert contains the label `send_managed_notification`, it is not routed to PagerDuty.

The former routing rule is placed in its own dedicated routing block that routes to the OCM Agent Operator. If CAMO 

The latter routing rule is always present in the Pagerduty routing block regardless of whether an OCM Agent is running on the cluster or not. 

## Pre-PR testing

This change has been successfully tested on a Production cluster to verify that it does not impact existing alerting. The following test scenarios were trialled:

- Created a new Critical severity alert with a "send_managed_notification" label of "true"; it was routed to OCM Agent but not to PagerDuty.
- Created a new Info severity alert with a "send_managed_notification" label of "true"; it was routed to OCM Agent but not to PagerDuty.
- Triggered a Critical severity alert that did not have a "send_managed_notification" label; it was routed to PagerDuty as expected.

Additional unit tests have been added to CAMO to cover the functionality that has been added.

## Implications for production deployment

OCM Agent Operator is not currently deployed to Production clusters.

There is no danger of this CAMO PR being included in a CAMO promotion ahead of time, as the `ocm-agent` ConfigMap will not exist on the clusters. CAMO's behaviour will therefore be unchanged.

## Documentation

A SOP covering the process of how to use CAMO to automate OCM Agent handling of an alert will be written as part of [OSD-10160](https://issues.redhat.com/browse/OSD-10160) prior to the Production rollout of this capability.

[0] https://github.com/openshift/ocm-agent-operator

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -3,6 +3,7 @@ package secret
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"os"
 	"strconv"
 
@@ -72,6 +73,25 @@ const (
 
 	// global config for PagerdutyURL
 	pagerdutyURL = "https://events.pagerduty.com/v2/enqueue"
+
+	// anything routed to "ocmagent" will not alert/notify SREP and will be handled by OCM Agent
+	receiverOCMAgent = "ocmagent"
+
+	// alert label used for identifying OCM Agent-bound alerts
+	managedNotificationLabel = "send_managed_notification"
+
+	// service name for the OCM Agent service
+	ocmAgentService = "ocm-agent"
+	// namespace for the OCM Agent service
+	ocmAgentNamespace = "openshift-ocm-agent-operator"
+	// path for the OCM Agent alertmanager receiver webhook
+	ocmAgentWebhookPath = "/alertmanager-receiver"
+
+	// configmap name for OCM agent configuration
+	cmNameOcmAgent = "ocm-agent"
+
+	// OCM Agent configmap key for service URL
+	cmKeyOCMAgent = "serviceURL"
 )
 
 var defaultNamespaces = []string{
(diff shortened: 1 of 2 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -260,6 +283,35 @@ func createPagerdutyRoute(namespaceList []string) *alertmanager.Route {
 	}
 }
 
+// createOCMAgentRoute creates an AlertManager Route for OcmAgent in memory.
+func createOCMAgentRoute() *alertmanager.Route {
+	return &alertmanager.Route{
+		Receiver: receiverOCMAgent,
+		Continue: true,
```
dofinn, line 290: Should this be false?

pkg/controller/secret/secret_controller.go
```diff
@@ -260,6 +283,35 @@ func createPagerdutyRoute(namespaceList []string) *alertmanager.Route {
...
+		Match:    map[string]string{managedNotificationLabel: "true"},
+		RepeatInterval: "10m",
+	}
+}
+
+// createOCMAgentReceiver creates an AlertManager Receiver for OCM Agent in memory.
+func createOCMAgentReceiver(ocmAgentURL string) []*alertmanager.Receiver {
+	if ocmAgentURL == "" {
+		return []*alertmanager.Receiver{}
```
dofinn, line 299: should this be handled explicity with error checking elsewhere? or devnull?

pkg/controller/secret/secret_controller.go
mrbarge: It ultimately doesn't matter as the same alerts get nullified in the pagerduty routing block, but I can't think of any danger of setting it to false either. I'll test this and update if AM's behaviour remains as expected.

PR #222, opened 2022-04-10
OSD-10936 Use a proxy for DMS/PD comms, if configured

mrbarge:
This PR adds the ability for CAMO to configure the `http_config` section of PagerDuty and DeadMansSnitch receivers to use a cluster-wide proxy if a HTTPS Proxy is found to be configured on a cluster.

This change is needed to handle scenarios where a cluster blocks all Internet-egress traffic that does not traverse via a proxy; in these situations comms to DMS/PD will be impacted as they do not travel via the proxy by default.

An additional permission has been added to CAMO's ClusterRole to be able to read the cluster's `proxy` resource.

Tests have been updated to check and verify the presence of the proxy in the resulting configs.

This change has been successfully tested in staging to verify that
- DMS and PD comms via a proxy configured in this manner are successful
- clusters without a proxy are not impacted

Refs [OSD-10936](https://issues.redhat.com//browse/OSD-10936)

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -321,7 +321,7 @@ func createOCMAgentReceiver(ocmAgentURL string) []*alertmanager.Receiver {
 }
 
 // createPagerdutyConfig creates an AlertManager PagerdutyConfig for PagerDuty in memory.
-func createPagerdutyConfig(pagerdutyRoutingKey, clusterID string) *alertmanager.PagerdutyConfig {
+func createPagerdutyConfig(pagerdutyRoutingKey, clusterID string, clusterProxy string) *alertmanager.PagerdutyConfig {
 	detailsMap := map[string]string{
 		"alert_name":   `{{ .CommonLabels.alertname }}`,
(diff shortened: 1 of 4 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -410,14 +412,25 @@ func createWatchdogReceivers(watchdogURL string) []*alertmanager.Receiver {
...
 }
 
+// createHttpConfig creates a HttpConfig used for receivers that can accept that configuration
+func createHttpConfig(clusterProxy string) alertmanager.HttpConfig {
+	if clusterProxy == "" {
+		return alertmanager.HttpConfig{}
+	}
+	return alertmanager.HttpConfig{
+		ProxyURL:  clusterProxy,
+		TLSConfig: alertmanager.TLSConfig{},
```
dofinn, line 422: do we need to fetch the CA bundle for the proxy and populate it here?

mrbarge: Based on my tests so far, as long as a cluster owner follows our procedure for adding their custom CA to the system CA truststore, it automatically gets included in the AM container's trusted CA bundle and no special handling needs to happen outside of that.

mrbarge: To specifically address your question here, AlertManager has a ConfigMap `alertmanager-trusted-ca-bundle` which has the `inject-trusted-cabundle` label, therefore, when a custom CA Bundle is added to the cluster, it will automatically be added to the AlertManager CA bundle by the cluster network operator. Therefore no CA bundle should need to be specified in the AlertManager config.

PR #223, opened 2022-04-18
Correcting indentation for config portion of olm template

blrm:
The config portion of the subscription definition in the OLM template wasn't indented far enough, so that portion of the subscription was missing when deployed. Adding that indentation back so that the FEDRAMP variable is set in the subscription object.


cblecker (approved):
/lgtm
/approve

PR #225, opened 2022-05-04
adding make-it-critical PD receiver, adding rules to accomodate OSD-11298

macgregor:
- added make-it-critical receiver to override severity of alerts to critical
- added rules to make MachineWithNoRunningPhase and MachineWithoutValidNode for "master" nodes critical rather than warning

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -155,11 +155,19 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 func createPagerdutyRoute(namespaceList []string) *alertmanager.Route {
 	// order matters.
 	// these are sub-routes.  if any matches it will not continue processing.
-	// 1. route anything we want to silence to "null"
-	// 2. route anything that should be a warning to "make-it-warning"
-	// 3. route anything that should be an error to "make-it-error"
-	// 4. route anything we want to go to PD
+	// 1. route anything we consider critical to "make-it-critical"
+	// 2. route anything we want to silence to "null"
+	// 3. route anything that should be a warning to "make-it-warning"
+	// 4. route anything that should be an error to "make-it-error"
+	// 5. route anything we want to go to PD
 	pagerdutySubroutes := []*alertmanager.Route{
+
+		// https://issues.redhat.com/browse/OSD-11298
+		// indications that master nodes have been terminated should be critical
+		// regex tests: https://regex101.com/r/Rn6F5A/1
+		{Receiver: receiverMakeItCritical, MatchRE: map[string]string{"name": "^.+-master-[123]$"}, Match: map[string]string{"alertname": "MachineWithoutValidNode", "namespace": "openshift-machine-api"}},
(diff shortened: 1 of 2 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -268,6 +271,10 @@ func createPagerdutyRoute(namespaceList []string) *alertmanager.Route {
 
 		// https://issues.redhat.com/browse/OSD-10485
 		{Receiver: receiverMakeItWarning, Match: map[string]string{"alertname": "etcdHighNumberOfFailedGRPCRequests", "namespace": "openshift-etcd"}},
+
+		// https://issues.redhat.com/browse/OSD-11298
+		{Receiver: receiverMakeItCritical, MatchRE: map[string]string{"name": ".*master.*"}, Match: map[string]string{"alertname": "MachineWithoutValidNode", "namespace": "openshift-machine-api"}},
+		{Receiver: receiverMakeItCritical, MatchRE: map[string]string{"name": ".*master.*"}, Match: map[string]string{"alertname": "MachineWithNoRunningPhase", "namespace": "openshift-machine-api"}},
```
macgregor, line 277: could match more specifically if we are worried about some node names like "mastercard-worker-1":
```
.*-master-[123]
```

macgregor, line 277: I've also seen these alerts that contain a number of these alerts aggregated together ([for example](https://redhat.pagerduty.com/alerts/Q13AQ8A1QJUUK3)). I assume we dont need to worry about whatever is aggregating the alerts messing up our override route?

dofinn: +1 - lets be as precise as possible with regex. something like `.+-master-[123]` ?

dofinn: the routing implimentaiton should be still be the same.

dofinn:
looks good bar few comments - when addressed could you please validate config/alert routing with https://www.prometheus.io/webtools/alerting/routing-tree-editor/ ?

pkg/controller/secret/secret_controller.go
macgregor: changed to `^.+-master-[123]$` and added a comment with link to regex test cases: https://regex101.com/r/Rn6F5A/1

macgregor:
> looks good bar few comments - when addressed could you please validate config/alert routing with https://www.prometheus.io/webtools/alerting/routing-tree-editor/ ?

Is there a way to do this prior to deploying the operator to stage/int/local? My plan was to deploy it to stage, create a cluster, terminate a master to see if the alert fires as critical, then terminate a non-master to see if the alert fires as warning.

dofinn:
> > looks good bar few comments - when addressed could you please validate config/alert routing with https://www.prometheus.io/webtools/alerting/routing-tree-editor/ ?
> 
> Is there a way to do this prior to deploying the operator to stage/int/local? My plan was to deploy it to stage, create a cluster, terminate a master to see if the alert fires as critical, then terminate a non-master to see if the alert fires as warning.

you can create a stage cluster, set its CAMO deployment to 0 via its CSV and run the operator locally. You could then either delete a master or copy the config it produced and use the promtool on the web and assess by providing an example payload (which we have from pagerduty)

PR #232, opened 2022-07-06
fix: duplicate FluentdNodeDown route

ninabauer:
[OSD-12353](https://issues.redhat.com//browse/OSD-12353)

Logging alerts without the SRE suffix were silenced in https://github.com/openshift/configure-alertmanager-operator/pull/230

FluentdNodeDown is still alerting because it is in the config twice. This routes the second FluentdNodeDown to null as well.


fahlmant (approved):
/approve
/lgtm

PR #233, opened 2022-07-11
set InsecureSkipVerify to true

wanghaoran1988:


dofinn:
@wanghaoran1988 is there a JIRA or explanation regarding this change?

NautiluX:
@reetika-vyas this is basically reverting PR #231 please have a look when you're back.
Haoran saw cluster-ready checks fail after `InsecureSkipVerify` was set to `false`
Card for the original change was OSD-11443 although I'm not sure how the card relates to changing `insecureSkipVerify`.

PR #235, opened 2022-07-20
OSD-12234 Redact client_url for FedRAMP

mjlshen:
It contains the user-supplied cluster name, which cannot leave the boundary due to compliance reasons

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -367,23 +367,28 @@ func createPagerdutyConfig(pagerdutyRoutingKey, clusterID string, clusterProxy s
 		"resolved":     `{{ template "pagerduty.default.instances" .Alerts.Resolved }}`,
 		"cluster_id":   clusterID,
 	}
+	clientURL := `{{ template "pagerduty.default.clientURL" . }}`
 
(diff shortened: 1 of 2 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -374,6 +374,18 @@ func createPagerdutyConfig(pagerdutyRoutingKey, clusterID string, clusterProxy s
...
 		detailsMap["firing"] = ``
 		detailsMap["link"] = ``
+
+		return &alertmanager.PagerdutyConfig{
+			NotifierConfig: alertmanager.NotifierConfig{VSendResolved: true},
+			RoutingKey:     pagerdutyRoutingKey,
+			Severity:       `{{ if .CommonLabels.severity }}{{ .CommonLabels.severity | toLower }}{{ else }}critical{{ end }}`,
+			// Defaults to '{{ template "pagerduty.default.clientURL" . }}' when empty, however this contains
+			// the cluster name which is sensitive information for FedRAMP, so setting it to "ROSA"
+			ClientURL:   "ROSA",
```
cblecker, line 384: instead of duplicating the whole config, can we set a variable to the default, and then override it for FR?

PR #238, opened 2022-09-05
OSD-12090 Update operator-sdk to 1.21

rendhalver:
Update configure-alertmanager-operator to use operator-sdk v1.21


rendhalver:
This is still WIP
/hold

controllers/secret_controller.go
```diff
@@ -98,53 +112,112 @@ var defaultNamespaces = []string{
...
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *SecretReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	client := mgr.GetClient()
+
+	r.Readiness = &readiness.Impl{Client: client}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.ConfigMap{}).
```
mrbarge, line 218: With this being a _secret_ controller it may be more correct to have this as a controller `For` Secrets (its primary resource) which also `Watches` ConfigMaps, rather than the other way around.

rendhalver: Thanks Matt. That does make more sense.
I will get that updated.

mrbarge: Thanks! The fix looks good.

controllers/secret_controller.go
```diff
@@ -98,53 +112,112 @@ var defaultNamespaces = []string{
...
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *SecretReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	client := mgr.GetClient()
+
+	r.Readiness = &readiness.Impl{Client: client}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Secret{}).
+		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{}).
```
bmeng, line 219: This should be replaced by `For` as well?

pkg/k8sutil/k8sutil.go
```diff
@@ -0,0 +1,82 @@
+package k8sutil
```
bmeng, line 1: This package used by `pkg/metrics/service.go` only. To get the operator name and namespace.
And those to functions can be replaced by const `config.OperatorName` and `config.OperatorNamespace`

So that, we do not need to introduce this package at all.

main.go
```diff
@@ -0,0 +1,168 @@
...
+	log.Info(fmt.Sprintf("Go Version: %s", runtime.Version()))
+	log.Info(fmt.Sprintf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
+}
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(configv1.AddToScheme(scheme))
+	utilruntime.Must(monitoringv1.AddToScheme(scheme))
+
+	utilruntime.Must(managedopenshiftiov1alpha1.AddToScheme(scheme))
```
bmeng, line 68: Is this needed? As there is nothing to add here?

main.go
```diff
@@ -0,0 +1,168 @@
...
+var log = logf.Log.WithName("cmd")
+
+func printVersion() {
+	log.Info(fmt.Sprintf("Go Version: %s", runtime.Version()))
+	log.Info(fmt.Sprintf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
+}
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(configv1.AddToScheme(scheme))
```
bmeng, line 65: Better to use `configv1.Install(scheme)` as the `AddToScheme` is deprecated.

main.go
```diff
@@ -0,0 +1,168 @@
...
+
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:                 scheme,
+		MetricsBindAddress:     fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		Port:                   9443,
+		HealthProbeBindAddress: probeAddr,
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "29fea974.configurealertmanager.managed.openshift.io",
```
bmeng, line 93: This `LeaderElectionID` is useless, as we go with the `Leader for life` approach with below 
```
err = leader.Become(context.TODO(), "name")
```

main.go
```diff
@@ -0,0 +1,168 @@
...
+			log.Info("Created ServiceMonitor")
+		}
+	}
+
+	log.Info("Starting prometheus metrics.")
+	if err := operatormetrics.StartMetrics(); err != nil {
+		log.Error(err, "Failed to start metrics service")
+		os.Exit(1)
+	}
+
```
bmeng, line 162:
```

	// Check for fedramp clusters
	err = operatorconfig.SetIsFedramp()
	if err != nil {
		log.Error(err, "Failed to get fedramp value")
		os.Exit(1)
	}
	if operatorconfig.IsFedramp() {
		log.Info("Running in fedramp environment.")
	}
```

This block to check the fedRamp value is missed before we start the manager

controllers/secret_controller.go
bmeng: cc @mrbarge I see you discussed about this above, but I am not sure what was  requested/changed.

mrbarge: You can only use `For` once in a controller. The only available option here is to do a `For` of the primary resource and a `Watches` of the secondary ones. The current usage is the best option available I think.

My other request concerned an earlier version of the code which was doing a `For` on `ConfigMap`, which is not the primary resource.

main.go
mrbarge: Its been moved higher in the code.

controllers/secret_controller.go
bmeng: ack, so it should leave as is

main.go
mrbarge: Given there is no API at all, we can perhaps explore removing the contents of the entire `api` directory entirely (but put a `.gitkeep` in there similar to what MNMO has done, so that the directory is retained to prevent breakage of OSDK or boilerplate), and it can be removed from here at the same time.

bmeng: I am thinking almost the same, if we do not need to add scheme here, so maybe the things under `apis/` can be removed as well. Just not sure how would it impact the operator.

But it is a bit weird that the operator does not have new api resource defined.

main.go
bmeng: overlooked... please ignore this one. Pete.

pkg/k8sutil/k8sutil.go
rendhalver: Ah. I added this to import the old code that was removed from the controller.

Thanks for the suggestion.
I would be happy to switch to a different solution.
I will see if I can work out how to do that.

rendhalver: I can basically replace the variable setup in service.go with this correct?
```
	operatorName := config.OperatorName
	namespace := config.OperatorNamespace
```

bmeng: Yeah, I think it would work the same. The current `GetOperatorNamespace()` or `GetOperatorName()` just return the same value as `config.OperatorNamespace` and `config.OperatorName`

rendhalver: I have made that change locally and the tests are passing.
I will push that up and the rest of my changes.

Thanks for the suggestion. :)

main.go
rendhalver: Thanks for spotting that.
Fixed in latest change.

main.go
rendhalver: Sounds good. I have made this change too.

main.go
rendhalver: I was wondering what to do about this too.
Thanks for the suggestions I will make those changes and see how it goes.

mrbarge:
Full functional testing of this change has been completed and documented in OSD-12976
/lgtm

cblecker:
/lgtm
/approve
/hold

@rendhalver Feel free to remove hold when you're ready for this to merge.

PR #246, opened 2022-11-04
OSD-11509 gracefully handle servicemonitor creation if service exists

mrbarge:
# What is this PR doing

Ensures that the presence of a `ServiceMonitor` occurs if the `Service` exists.

# Why is it needed

This is a bug fix for [OSD-11509](https://issues.redhat.com//browse/OSD-11509), which recently came up again in conversation.

If the CAMO `ServiceMonitor` goes missing, a restart of the CAMO pod won't result in its recreation if the `Service` exists. You'd need to delete the `Service` in order for both to then get recreated. This is just a little quirk in the existing code flow that is easily handled by checking for the `AlreadyExists` error.

The `go.mod` and `go.sum` changes are the result of freshly running `go mod tidy`. On a fresh clone of the repo my build was originally complaining about the current state of these.


main.go
```diff
@@ -138,7 +139,7 @@ func main() {
 
 	sm := operatormetrics.GenerateServiceMonitor(s)
 	err = mgr.GetClient().Create(context.TODO(), s)
-	if err != nil {
+	if err != nil && !apierrors.IsAlreadyExists(err) {
```
mjlshen, line 142: This makes is slightly misleading when we
```
log.Info("Created Service")
```
but I don't think it would cause confusion - maybe we could log that we `ensured metrics service exists` or something. Not a big deal for me

mjlshen (approved):
/lgtm

main.go
cblecker: Yeah +1 to cleaning up the logging in this function while we are here

mrbarge: Agree! I've given that whole section a bit of a tidy-up now, thanks for calling it out.

mjlshen (approved):
/lgtm

Can't do any more than that :D

PR #247, opened 2022-11-10
OSD-13721 - Inhibit api-ErrorBudgetBurn when KubeAPIErrorBudgetBurn is firing

tnierman:
Completes: https://issues.redhat.com/browse/OSD-13721

---

These changes add an `inhibitRule` which prevents `api-ErrorBudgetBurn` alerts from firing when a `KubeAPIErrorBudgetBurn` alert has already been triggered.

An inverse rule was not created because `KubeAPIErrorBudgetBurn` alerts are based on the kube-apiservers' internal metrics while `api-ErrorBudgetBurn` alerts are based on the reachability of the kube-apiservers. Firing a `KubeAPIErrorBudget` burn after an `api-ErrorBudgetBurn` is, at this time, believed to indicate a separate issue with the control plane which needs to be addressed individually.

### DIFF
--- a/controllers/secret_controller.go
+++ b/controllers/secret_controller.go
@@ -692,6 +692,18 @@ func createAlertManagerConfig(pagerdutyRoutingKey, watchdogURL, ocmAgentURL, clu
 					"alertname": "ElasticsearchClusterNotHealthy",
 				},
 			},
+			// https://issues.redhat.com/browse/OSD-13721
+			{
+				Equal: []string{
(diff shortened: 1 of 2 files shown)


rendhalver:
Will this affect our SLOs if the `api-ErrorBudgetBurn` alert is inhibited and not acknowledged.
Should this be the other way around if that is the case?

NautiluX (commented):
/lgtm

PR #248, opened 2022-11-16
Silence Insights operator down alert in FedRAMP environment

katherinelc321:
This insights operator is not available in the Fedramp environment yet and it's paging SREPs unnecessarily. This PR silence the ClusterOperatorDown alert for now. It will need to be removed once insights is made available in the environment.
It's been tested in a fedramp cluster successfully
https://issues.redhat.com/browse/OSD-13685

### DIFF
--- a/controllers/secret_controller.go
+++ b/controllers/secret_controller.go
@@ -388,7 +388,6 @@ func createPagerdutyRoute(namespaceList []string) *alertmanager.Route {
 		pagerdutySubroutes = append(pagerdutySubroutes,
 			[]*alertmanager.Route{
 				{Receiver: receiverNull, Match: map[string]string{"alertname": "ClusterOperatorDown", "name": "insights"}},
-				{Receiver: receiverNull, Match: map[string]string{"alertname": "ClusterOperatorDegraded", "name": "insights"}},
 			}...,
 		)
 	}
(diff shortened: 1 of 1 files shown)


controllers/secret_controller.go
```diff
@@ -382,6 +382,15 @@ func createPagerdutyRoute(namespaceList []string) *alertmanager.Route {
 		{Receiver: receiverMakeItWarning, Match: map[string]string{"severity": "critical", "namespace": "openshift-deployment-validation-operator"}},
 	}
 
+	// Silence insights in FedRAMP until its made available in the environment
+	// https://issues.redhat.com/browse/OSD-13685
+	if config.IsFedramp() {
+		pagerdutySubroutes = append(pagerdutySubroutes,
+			[]*alertmanager.Route{
+				{Receiver: receiverNull, Match: map[string]string{"alertname": "ClusterOperatorDown", "name": "insights"}},
+			}...)
```
mjlshen, line 391: Since there's only one exception I think we should just do:
```go
	if config.IsFedramp() {
		pagerdutySubroutes = append(pagerdutySubroutes,
			&alertmanager.Route{Receiver: receiverNull, Match: map[string]string{"alertname": "ClusterOperatorDown", "name": "insights"}},
		)
	}
```

mjlshen (changes requested):

cblecker:
/approve
/retitle Silence Insights operator down alert in FedRAMP environment

PR #251, opened 2022-12-07
Linting fixes

cblecker:
This fixes a couple of linting errors:
- ignore G114 for the metrics endpoint as we don't need to worry about enforcing timeouts
- move to `os.ReadFile` instead of `ioutil.ReadFile` which has been deprecated


fahlmant (approved):
/lgtm
/approve

PR #258, opened 2023-01-17
Only configure DMS if PD is being configured

mrbarge:
## What does this PR do?

- Alters the logic of the `secret` controller to only create a Dead Man's Snitch watchdog route/receiver if there is also a Pagerduty receiver needing to be configured.
- Includes some logging in the `createAlertManagerConfig` function, because more visibility into what the operator is doing is personally preferred.
- Reworks some of the test case helper functions to be position-independent when validating routes/receivers in the Alertmanager config.

## Why?

This was identified via incident [OHSS-17569](https://issues.redhat.com//browse/OHSS-17569) as an opportunity for improvement (and it was mistakenly believed the operator already behaved like this).

As PagerDuty and DMS are both required for monitoring, the lack of PagerDuty being configured on a cluster is an event that should rely on the other monitoring (in this case, DMS's watchdog) to raise an alarm to SRE that the cluster is "flying blind", as it were. By not configuring the watchdog, we trigger this behaviour.

## Jira

Refs [OSD-14346](https://issues.redhat.com//browse/OSD-14346)

## Docs

- The project `README` has been updated.
- The associated CHGM SOP will be updated to include a 'check PD/DMS in AlertManager config' section after merge of this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@
 ## Summary
 The Configure Alertmanager Operator was created for the OpenShift Dedicated platform to dynamically manage Alertmanager configurations based on the presence or absence of secrets containing a Pager Duty RoutingKey and [Dead Man's Snitch](https://deadmanssnitch.com) URL. When the secret is created/updated/deleted, the associated Receiver and Route will be created/updated/deleted within the Alertmanager config.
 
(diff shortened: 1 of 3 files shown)


cblecker:
/lgtm
/approve

Sorry for the delay on review, but thanks for this one @mrbarge!

PR #259, opened 2023-01-24
Silencing NodeFilesystemAlmostOutOfSpace alert as per OSD-14857

ravitri:
### What type of PR is this?
_(bug)_

### What this PR does / why we need it?
Based on [OSD-14857](https://issues.redhat.com/browse/OSD-14857) and [OCPBUGS-6577](https://issues.redhat.com/browse/OCPBUGS-6577), the NodeFilesystemAlmostOutOfSpace alert wrongly fires if a tmpfs mount even with only 4K size is 100% used. This is not actionable by SRE. This issue is fixed in 4.13 for cluster-monitoring-operator but yet to be fixed in older versions. PR https://github.com/openshift/managed-cluster-config/pull/1436 introduces NodeFilesystemAlmostOutOfSpaceSRE alert which would help reduce some noise till the backports and new release is released and used by the customer(s). Thus, we can silence the NodeFilesystemAlmostOutOfSpace alert for now.

4.13+ Fix - [Link](https://github.com/openshift/cluster-monitoring-operator/blob/release-4.13/assets/node-exporter/prometheus-rule.yaml#L68-L82)

### Which Jira/Github issue(s) this PR fixes?

_Fixes #_ [OSD-14857](https://issues.redhat.com/browse/OSD-14857)

### DIFF
--- a/controllers/secret_controller.go
+++ b/controllers/secret_controller.go
@@ -383,6 +383,9 @@ func createPagerdutyRoute(namespaceList []string) *alertmanager.Route {
 
 		// https://issues.redhat.com/browse/OSD-14071
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "MultipleDefaultStorageClasses", "namespace": "openshift-cluster-storage-operator"}},
+
+		// https://issues.redhat.com/browse/OSD-14857
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "NodeFilesystemAlmostOutOfSpace", "severity": "critical"}},
 	}
 
 	// Silence insights in FedRAMP until its made available in the environment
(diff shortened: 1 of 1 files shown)


controllers/secret_controller.go
```diff
@@ -383,6 +383,9 @@ func createPagerdutyRoute(namespaceList []string) *alertmanager.Route {
 
 		// https://issues.redhat.com/browse/OSD-14071
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "MultipleDefaultStorageClasses", "namespace": "openshift-cluster-storage-operator"}},
+
+		// https://issues.redhat.com/browse/OSD-14857
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "NodeFilesystemAlmostOutOfSpace", "severity": "critical"}},
```
cblecker, line 388: Can we just use `MatchRE` to silence the alerts with a label matching the mountpoints we want to ignore? That way we don't have to create a whole new alert,

cblecker (changes requested):

controllers/secret_controller.go
ravitri: @cblecker - Nice suggestion! Thank you. Yup, I have added `MatchRE` to match the problematic mountpoint to be ignored. I'll close the MCC PR then for new alert.

ravitri:
/unhold

Removing hold for further review/approval

PR #267, opened 2023-02-24
OSD-14149 : Modified the config to enable CannotRetrieveUpdates alert…

squirrd:
This alert was disabled due to [bug 2109374](https://bugzilla.redhat.com/show_bug.cgi?id=2109374) in [OSD-6327](https://issues.redhat.com/browse/OSD-6327)

If this alert fires it means the cluster version operator has not retrieved updates in a reasonable amount of time.  This PR is to re-enable the alert as critical.

The impact to the cluster is that the CVO is not able to refresh the available updates successfully. This impacts the customer schedules because the version that the customer wishes to upgrade to is not in the list of available upgrades, the MUO does not proceed with the upgrade and it is eventually cancelled.

When this happens, we tend to get OHSS queries from the customer asking why an upgrade was canceled.


a7vicky:
@squirrd PR need rebase to resolve merge conflict.

squirrd:
This PR is closed as the rebasing after the latest merge is too difficult. The new PR is #268

PR #279, opened 2023-04-25
Adds supression to ClusterOperatorDown on monitoring operator

iamkirkbater:
Suppresses the ClusterOperatorDown alert for the monitoring operator. Allows us to break this alert up with https://github.com/openshift/managed-cluster-config/pull/1641 and eventually send the multiple prometheus instances to OAO to send servicelogs.


mjlshen:
What do you think about letting this alert stay for "a period of time", which would result in a double-ping for SREs when CMO is degraded to a second monitoring stack with regards to the comment in https://github.com/openshift/managed-cluster-config/pull/1641/files#r1177435392 since most of our alert analysis tools I'm aware of are PD-based:

> Ultimately I just harbor a little concern about the potential of accidentally masking genuine cluster-monitoring-operator issues due to the presence of an innocuous Prometheus that's not causing any trouble.

I think this would let us gather data to conclude whether or not there are false positive/negatives out there.

PR #280, opened 2023-05-09
Suppresses NodeFileDescriptorLimit Alerts in lieu of other SRE-flavored alerts for the same.

iamkirkbater:
Routes NodeFileDescriptorLimit alerts to the null receiver since there are now two new alerts, one for ControlPlane/Infra nodes and one for Worker nodes, and the worker nodes one is set to automatically send an SL to the customer, skipping SRE.

https://issues.redhat.com/browse/OSD-12379


NautiluX (commented):
/lgtm

PR #284, opened 2023-06-07
osde2e: initial commit

jbpratt:
- Boilerplate: Update to d3612328c972f49103b8fd1be49cd8b0a387c980
- osde2e: initial commit

[SDCICD-1022](https://issues.redhat.com//browse/SDCICD-1022)


jbpratt:
depends on https://github.com/openshift/osde2e-common/pull/3

jbpratt:
This is ready to go

/cc @ryankwilliams @dustman9000

ryankwilliams (approved):

go.mod
```diff
@@ -1,20 +1,23 @@
 module github.com/openshift/configure-alertmanager-operator
```
cblecker, line 1: Why are we bumping the golang version and all these dependencies in this PR?

go.mod
```diff
@@ -1,20 +1,23 @@
...
+	github.com/prometheus/common v0.43.0
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.24.3
-	k8s.io/apimachinery v0.24.3
-	k8s.io/client-go v0.24.2
-	sigs.k8s.io/controller-runtime v0.12.3
+	k8s.io/api v0.27.2
+	k8s.io/apimachinery v0.27.2
+	k8s.io/client-go v0.27.2
+	sigs.k8s.io/controller-runtime v0.15.0
```
cblecker, line 20: These aren't even going to have an effect because they are replaced at the bottom

go.mod
jbpratt: This PR is adding the initial refactored suite of tests from osde2e. This is a requirement of using the framework we are writing.

Why are we staying on an end of life version of Go?

go.mod
jbpratt: they are bumped when we run go mod tidy because of the dependencies, it does not matter if they are replaced. Things work as intended within the tests regardless of the pinning below.

PR #286, opened 2023-07-03
Sends NodeMemoryUsage to null receiver

iamkirkbater:
Since this is just a high-watermark alert at 85% memory usage, it provides little actual value, and should be silenced as there's nothing that should be done here.


iamkirkbater:
Closing this - this is not a product alert and is a custom alert the customer added bypassed webhooks to add to the cluster.

PR #291, opened 2023-08-03
[OSD-16014] suppress the cluster operator alert HAProxyDown

bmeng:
Suppress the HAProxyDown alert from the cluster ingress operator.
Split it do HAProxyDownSRE (default IC) and HAProxyDownSRE (non-default IC), so we can be alerted on the default IC only when the ingresscontroller can be managed by CU.

### DIFF
--- a/controllers/secret_controller.go
+++ b/controllers/secret_controller.go
@@ -35,6 +35,7 @@ import (
 
 	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
+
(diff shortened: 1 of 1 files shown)


hectorakemp (approved):

cblecker:
/approve
/hold

@bmeng Holding to ensure that the alerts in MCC are merged and tested first. Please also ensure the MCC changes are promoted first before promoting CAMO

bmeng:
/unhold
as the MCC part promoted to production

PR #295, opened 2023-09-13
silence PodDisruptionBudgetLimit completely

maorfr:
part of https://issues.redhat.com/browse/OSD-14158

silencing this alert in order to automate SL via MCC: https://github.com/openshift/managed-cluster-config/pull/1835


cblecker:
/lgtm
/approve
/hold

Holding as this should only merge after the superseding alert has been merged/promoted.

maorfr:
for the sake of reducing time zone waiting time, we can also let this merge and not promote until the dependency is promoted.

PR #298, opened 2023-10-06
feat(OSD-19000): mute critical KubePersistentVolumeFillingUp as it is handled by OAO

typeid:
Disable Critical KubePersistentVolumeFillingUp as it is is handled by OAO. 
To allow OAO handling it directly, we now define the upstream expression directly in MCC instead of counting active alerts for this (https://github.com/openshift/managed-cluster-config/pull/1895) . 

Fixes https://issues.redhat.com/browse/OSD-19000


mjlshen (approved):
/lgtm

cblecker (approved):
/lgtm

PR #302, opened 2023-10-27
OSD-19439: Route CertificateIsAboutToExpire to nullReceiver

typeid:
This is a tentative PR to tackle https://issues.redhat.com/browse/OSD-19439. 

`CertificateIsAboutToExpire` is an alert coming from the community [cert-utils-operator](https://github.com/redhat-cop/cert-utils-operator) - we temporarily relaxed our webhook to allow creating this alerting rule in `openshift-monitoring`, see https://issues.redhat.com/browse/OSD-13909. 

From my understanding, we can just route this to `nullReceiver` anyway, as the alert will still be available for the customer in the console.


NautiluX (commented):
/lgtm

PR #309, opened 2023-12-01
OSD-19800: silence insights operator down

typeid:
**What?**

Silences the insights operator down alert for all clusters (fedramp & osd).

**Why?**

This is being replaced with a custom ClusterOperatorDown in MCC  with a longer down duration as requirement for it to trigger. See https://issues.redhat.com/browse/OSD-19800


NautiluX (commented):
/lgtm

PR #310, opened 2023-12-05
OSD-19310 adjusting CAMO for FedRAMP GoAlert

theautoroboto:
We were seeing issues where SREP was being alerted to customer namespace critical alerts.  After investigating the issue, this was due to the default parent receiver route being set to `goalert`.  Changing the default receiver to "null", any alerts that are not specifically defined will be delivered to "null" by default.

```
route:
  receiver: "null"
  group_by:
  - job
  continue: false
  routes:
  - receiver: "null"
    group_by:
    - alertname
    - severity
    continue: true
    routes:
```

To ensure we do not miss alerts for managed namespaces, we are creating a route for each managed-namespace for all severities of `critical`, `error`, and `warning` to the correct receivers.  As a note, we should plan to convert these from the deprecated `match` and `match_re` to `matchers`.
```
    - receiver: goalert-high
      match:
        exported_namespace: ""
        prometheus: openshift-monitoring/k8s
        severity: critical
      match_re:
        namespace: ^openshift-network-operator$
      continue: false
    - receiver: goalert-high
      match:
        exported_namespace: ""
        prometheus: openshift-monitoring/k8s
        severity: error
      match_re:
        namespace: ^openshift-network-operator$
      continue: false
    - receiver: goalert
      match:
        exported_namespace: ""
        prometheus: openshift-monitoring/k8s
        severity: warning
      match_re:
        namespace: ^openshift-network-operator$
      continue: false
```

### DIFF
--- a/controllers/secret_controller.go
+++ b/controllers/secret_controller.go
@@ -269,7 +269,7 @@ func createSubroutes(namespaceList []string, receiver receiverType) *alertmanage
 		receiverCritical = receiverGoAlertHigh
 		receiverError = receiverGoAlertHigh
 		receiverWarning = receiverGoAlertLow
-		receiverDefault = receiverGoAlertLow
+		receiverDefault = receiverNull
 	case Pagerduty:
 		receiverCommon = receiverPagerduty
 		receiverCritical = receiverMakeItCritical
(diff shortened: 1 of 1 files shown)


dustman9000:
Is there an OSD card tracking this issue/change?

mjlshen (approved):
/label tide/merge-method-squash
/lgtm

PR #323, opened 2024-05-10
OSD-23026: Update regex for master node names to account for CPMS

jewzaam:
In https://issues.redhat.com/browse/OSD-23026 found the regex for machine names was no longer valid with the introducction of CPMS. Impacts two alerts:

* MachineWithNoRunningPhase
* MachineWithoutValidNode 

And results in those alerts not having severity bumped to critical when sent to PD (they stay at warning) and are therefore not raised to on-call.

### DIFF
--- a/controllers/secret_controller.go
+++ b/controllers/secret_controller.go
@@ -297,8 +297,8 @@ func createSubroutes(namespaceList []string, receiver receiverType) *alertmanage
 		// https://issues.redhat.com/browse/OSD-11298
 		// indications that master nodes have been terminated should be critical
 		// regex tests: https://regex101.com/r/Rn6F5A/1
-		{Receiver: receiverCritical, MatchRE: map[string]string{"name": "^.+-master-[123]$"}, Match: map[string]string{"alertname": "MachineWithoutValidNode", "namespace": "openshift-machine-api"}},
-		{Receiver: receiverCritical, MatchRE: map[string]string{"name": "^.+-master-[123]$"}, Match: map[string]string{"alertname": "MachineWithNoRunningPhase", "namespace": "openshift-machine-api"}},
+		{Receiver: receiverCritical, MatchRE: map[string]string{"name": "^.+-master-.*[0-9]+$"}, Match: map[string]string{"alertname": "MachineWithoutValidNode", "namespace": "openshift-machine-api"}},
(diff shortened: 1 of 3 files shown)


controllers/secret_controller.go
```diff
@@ -297,8 +297,8 @@ func createSubroutes(namespaceList []string, receiver receiverType) *alertmanage
 		// https://issues.redhat.com/browse/OSD-11298
 		// indications that master nodes have been terminated should be critical
 		// regex tests: https://regex101.com/r/Rn6F5A/1
-		{Receiver: receiverCritical, MatchRE: map[string]string{"name": "^.+-master-[123]$"}, Match: map[string]string{"alertname": "MachineWithoutValidNode", "namespace": "openshift-machine-api"}},
-		{Receiver: receiverCritical, MatchRE: map[string]string{"name": "^.+-master-[123]$"}, Match: map[string]string{"alertname": "MachineWithNoRunningPhase", "namespace": "openshift-machine-api"}},
+		{Receiver: receiverCritical, MatchRE: map[string]string{"name": "^.+-master-.*[0-9]+$"}, Match: map[string]string{"alertname": "MachineWithoutValidNode", "namespace": "openshift-machine-api"}},
+		{Receiver: receiverCritical, MatchRE: map[string]string{"name": "^.+-master-.*[0-9]+$"}, Match: map[string]string{"alertname": "MachineWithNoRunningPhase", "namespace": "openshift-machine-api"}},
```
mjlshen, line 301:
```suggestion
		{Receiver: receiverCritical, MatchRE: map[string]string{"name": "^.+-master-.*[0-9]$"}, Match: map[string]string{"alertname": "MachineWithoutValidNode", "namespace": "openshift-machine-api"}},
		{Receiver: receiverCritical, MatchRE: map[string]string{"name": "^.+-master-.*[0-9]$"}, Match: map[string]string{"alertname": "MachineWithNoRunningPhase", "namespace": "openshift-machine-api"}},
```
curious about the `+`, is it needed?

mjlshen (commented):
Need to update the Dockerfiles in `build/` as well to get the latest image

```
FROM registry.access.redhat.com/ubi8/ubi-minimal:8.9-1161.1715068733
```

controllers/secret_controller.go
jewzaam: The + will ensure 1 or more number.  We don't need it today but we're hitting this bug because of assumptions of naming standard stability.  I figure hedging a little bit here is safe.

mjlshen (approved):
/lgtm

PR #342, opened 2024-10-01
OSD-22337 route some customer alerts to null receiver

tkong-redhat:
There are some customer defined alerts which are picked up by CAMO and routed to Red Hat pagerduty due to customer set some labels with Red Hat managed namespaces.

Adding some route rules to redirect those alerts to NULL receiver so that Red Hat SRE will not be paged.

Rule has been tested on a stg cluster.

### DIFF
--- a/controllers/secret_controller.go
+++ b/controllers/secret_controller.go
@@ -457,6 +457,13 @@ func createSubroutes(namespaceList []string, receiver receiverType) *alertmanage
 		// Route ClusterOperatorDown for insights to null receiver https://issues.redhat.com/browse/OSD-19800
 		// Also needs to be silenced for FedRAMP until its made available in the environment https://issues.redhat.com/browse/OSD-13685
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "ClusterOperatorDown", "name": "insights"}},
+		// Route some customer defined alerts to null receiver
+		// https://issues.redhat.com/browse/OSD-22337
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "memory-request"}},
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "cpu-request"}},
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "daemonset-availability"}},
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "sonarqube-ErrorBudgetBurn"}},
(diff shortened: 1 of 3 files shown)


controllers/secret_controller.go
```diff
@@ -457,6 +457,13 @@ func createSubroutes(namespaceList []string, receiver receiverType) *alertmanage
 		// Route ClusterOperatorDown for insights to null receiver https://issues.redhat.com/browse/OSD-19800
 		// Also needs to be silenced for FedRAMP until its made available in the environment https://issues.redhat.com/browse/OSD-13685
 		{Receiver: receiverNull, Match: map[string]string{"alertname": "ClusterOperatorDown", "name": "insights"}},
+		// Route some customer defined alerts to null receiver
+		// https://issues.redhat.com/browse/OSD-22337
+		{Receiver: receiverNull, Match: map[string]string{"alertname": "memory-request"}},
```
dustman9000, line 462: This is not a great long term solution to filter on alertname, is there a better way to identify these "user defined" alerts?

tkong-redhat: This is a short term solution. There is no easy way to identify those "user defined" alerts. We had several discussion about this. There is a doc about our finding https://docs.google.com/document/d/1OscbdlZ-aBuwY7YKJsU5URyDvfS9A4-VV4yBhpowfy4/edit#heading=h.i7a6b1h441g

tkong-redhat: We should use some better labels to distinguish between Red Hat defined alert and user defined alert. But that requires more discussion and nearly impact all the existing alerts. That's why we put a short term solution here.

xiaoyu74: > This is not a great long term solution to filter on alertname, is there a better way to identify these "user defined" alerts?

Agree, looks this PR’s current logic relies on matching `specific alertname values` to identify user-defined alerts and route them to a null. However, if any change in the alert name or new customer-defined alerts not explicitly added in this list can bypass the receiever.

probably use a label `user_alert=true` or something similar to match alerts instead of relying on alertnames?

tkong-redhat: TBH, I would prefer long term solution which is adding some common labels so we could easily distinguish user defined alerts and Red Hat defined alerts. That would be lovely and this makes our life easier. Unfortunately, we don't have this type of definition in any our docs, so there is no standard labels to be used ATM. It requires a lot of effort to discuss with BU and wider teams about what should be the labels used and make a agreement.

PR #350, opened 2024-11-19
config: enable OLM skip range

jbpratt:
to allow OLM to skip broken versions and continue upgrading


jbpratt:
/hold

this should not be merged yet

PR #390, opened 2025-02-24
Silencing ClusterNodeDown alert

ravitri:
This PR silences the `ClusterNodeDown` which is a custom alert that's not actionable.

For the time being there's isn't a quick and straightforward way to identify and deny the custom alerts yet (besides changing to a allowlist model) so will silence it individually for now. We can revert this change if a better longterm solution is identified separately.

Regards https://issues.redhat.com/browse/OSD-26543 JIRA


ravitri:
/hold

Adding hold to clear the tests

ravitri:
/close

For now closing this PR. We'll review the alert further and then disable it if required.

PR #403, opened 2025-04-07
run make boilerplate-update

jimdaga:


.ci-operator.yaml
```diff
@@ -1,4 +1,4 @@
 build_root_image:
   name: boilerplate
   namespace: openshift
-  tag: image-v6.0.1
+  tag: image-v7.0.0
```
joshbranham, line 4: I know this is just a draft, but we should update the operator to use Go 1.23 as well, since that is what comes in this CI image.

jbpratt:
this can be closed, #405 bumped BP!

PR #526, opened 2026-05-15
ROSAENG-1157 | Drop ClusterLogForwarderOutputErrorRate alerts

AlexSmithGH:
This alert was added here: https://github.com/openshift/cluster-logging-operator/pull/3137. This alert is not something SREP should be tasked with triaging, so silencing it here.

https://redhat.pagerduty.com/incidents/Q2LTARFJT6Z7XU - Triggered twice for two clusters owned by the same customer, and was related directly to their config. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated alert silencing configuration to properly handle cluster log forwarder output rate alerts in the logging namespace.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -8,7 +8,7 @@ COPY . .
 RUN make go-build
 
 ####
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.7-1778461551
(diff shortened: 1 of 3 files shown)


clcollins:
/lgtm

We have followup work for the stuff for rosa-log-forwarder planned, and we shouldn't be alerting for this across the entire fleet at all.

PR #553, opened 2026-05-28
Reapply "SREP-2513: Separate PD and CAD_PD secrets"

rolandmkunkel:
Separates the PD and CAD_PD secrets into distinct secrets (pd-secret and cad-pd-secret) instead of reading both keys from pd-secret.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed PagerDuty routing detection so CAD-specific routing is read from the correct secret when present and the cluster is ready; improved logging for secret presence and readiness states.

* **Tests**
  * Updated test setups to use a unified secret creation helper and validate the corrected detection and handling across initialization scenarios.

### DIFF
--- a/controllers/secret_controller.go
+++ b/controllers/secret_controller.go
@@ -76,6 +76,8 @@ const (
 
 	secretNamePD = "pd-secret"
 
+	secretNameCADPD = "cad-pd-secret" // #nosec G101
+
 	secretNameDMS = "dms-secret"
 
 	secretNameAlertmanager = "alertmanager-main"
@@ -191,6 +193,7 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 	switch request.Name {
 	case secretNameGoalert:
 	case secretNamePD:
+	case secretNameCADPD:
 	case secretNameDMS:
 	case secretNameAlertmanager:
 	case cmNameOcmAgent:
@@ -1066,6 +1069,7 @@ func (r *SecretReconciler) parseSecrets(reqLogger logr.Logger, secretList *corev
 	// Check for the presence of specific secrets.
 	goalertSecretExists := secretInList(reqLogger, secretNameGoalert, secretList)
 	pagerDutySecretExists := secretInList(reqLogger, secretNamePD, secretList)
+	cadPagerDutySecretExists := secretInList(reqLogger, secretNameCADPD, secretList)
 	snitchSecretExists := secretInList(reqLogger, secretNameDMS, secretList)
 
 	// If a secret exists, add the necessary configs to Alertmanager.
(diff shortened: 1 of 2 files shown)


controllers/secret_controller.go
```diff
@@ -1076,14 +1080,25 @@ func (r *SecretReconciler) parseSecrets(reqLogger logr.Logger, secretList *corev
...
 	} else {
 		reqLogger.Info("INFO: Pager Duty secret does not exist")
 	}
 
+	if cadPagerDutySecretExists {
+		reqLogger.Info("INFO: CAD Pager Duty secret exists")
+		if clusterReady {
+			reqLogger.Info("INFO: Cluster is ready; configuring CAD Pager Duty")
+			cadPagerdutyRoutingKey = readSecretKey(r, secretNameCADPD, namespace, secretKeyCADPD)
+		} else {
```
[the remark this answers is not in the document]
rolandmkunkel: The problem here lies in readSecretKey and not the new code added. Also, this code has been used in production for years. It also retries indirectly, because of the reconciliation loop. I will create a follow up ticket for this.

nephomaniac (commented):
Looks good overall. A couple suggestions below.

controllers/secret_controller.go
```diff
@@ -1076,14 +1080,25 @@ func (r *SecretReconciler) parseSecrets(reqLogger logr.Logger, secretList *corev
...
 			reqLogger.Info("INFO: Cluster is not ready; skipping Pager Duty configuration")
 		}
 	} else {
 		reqLogger.Info("INFO: Pager Duty secret does not exist")
 	}
 
+	if cadPagerDutySecretExists {
+		reqLogger.Info("INFO: CAD Pager Duty secret exists")
+		if clusterReady {
+			reqLogger.Info("INFO: Cluster is ready; configuring CAD Pager Duty")
```
nephomaniac, line 1093: Consider logging a warning if the secret exists but the key is empty — this would indicate a misconfigured secret vs a cluster that simply doesn't have CAD support:
```go
cadPagerdutyRoutingKey = readSecretKey(r, secretNameCADPD, namespace, secretKeyCADPD)
if cadPagerdutyRoutingKey == "" {
    reqLogger.Info("WARNING: CAD Pager Duty secret exists but configuration is empty, skipping CAD configuration")
}
```

controllers/secret_controller.go
```diff
@@ -1076,14 +1080,25 @@ func (r *SecretReconciler) parseSecrets(reqLogger logr.Logger, secretList *corev
...
+			cadPagerdutyRoutingKey = readSecretKey(r, secretNameCADPD, namespace, secretKeyCADPD)
+		} else {
+			reqLogger.Info("INFO: Cluster is not ready; skipping CAD Pager Duty configuration")
+		}
+	} else {
```
nephomaniac, line 1101: Would it be worth falling back to reading from `pd-secret` if `cad-pd-secret` doesn't exist? This would handle cases where the new secret hasn't been provisioned yet, allowing CAD routing to continue working via the existing pd-secret.
```go
} else if pagerDutySecretExists && clusterReady {
    reqLogger.Info("INFO: CAD Pager Duty secret not found, falling back to pd-secret")
    cadPagerdutyRoutingKey = readSecretKey(r, secretNamePD, namespace, secretKeyCADPD)
} else {
    reqLogger.Info("INFO: CAD Pager Duty secret does not exist")
}
```
Understand if you'd prefer this to fail explicitly without the secret — just wanted to flag as an option.

controllers/secret_controller.go
rolandmkunkel: Good suggestion, I've added it with a change of wording

controllers/secret_controller.go
rolandmkunkel: The new secret is required for a route (created with createCADPagerdutyRoute()) which sends automation alerts with the route_to_cad label directly to CAD without SRE involvement. These type of alerts are not meant to page SRE, and use a different escalation policy. If we added this fallback, they would page SRE, which is not what we want, as these are not actionable.

